PR #39, opened 2023-01-20
Clarify use of language codes in README

### LINKED ISSUE #30: Language tags are not always from ISO 639-1
This may be nitpicking, but here goes.

The top level [README](https://github.com/adbar/simplemma#supported-languages) states:

> The following languages are available using their [ISO 639-1 code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes):

But this isn't true in all cases. Currently supported three letter codes `enm` (Middle English) and `hbs` (Serbo-Croatian) are not from ISO 639-1 but from ISO 639-3.

Should this be changed to refer to [IETF BCP 47 language tags](https://en.wikipedia.org/wiki/IETF_language_tag) instead? A downside is that the Wikipedia page only lists some of the more common languages (for understandable reasons, since there are so many valid codes). In particular, Middle English and Serbo-Croatian are not listed there.

osma:
This little PR to the README explains that BCP 47 language tags are used, which means mostly using ISO 639-1 but if no such code exists, ISO 639-3 codes are used instead.

Fixes #30

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -190,7 +190,7 @@ The ``lang_detector()`` function returns a list of language codes along with sco
 Supported languages
 -------------------
 
-The following languages are available using their `ISO 639-1 code <https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes>`_:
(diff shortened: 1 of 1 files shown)


adbar:
Hi @osma, thanks for the PR, I'll push an updated version of the README first and then accept it.

PR #41, opened 2023-01-23
Sort dict before pickling

1over137:
Sorting is known to noticeably improve the efficiency of compression algorithms due to the ability for them to reuse dicts across a shorter distance. From my own tests it reduces dictionary size by ~10% completely for free.

### DIFF
--- a/simplemma/dictionary_pickler.py
+++ b/simplemma/dictionary_pickler.py
@@ -6,14 +6,15 @@ import re
 from pathlib import Path
 from typing import Dict, List, Optional
 
-from .constants import LANGLIST
-from .utils import levenshtein_dist
-
 try:
+    from .constants import LANGLIST
     from .rules import apply_rules
+    from .utils import levenshtein_dist
 # local error, also ModuleNotFoundError for Python >= 3.6
 except ImportError:  # pragma: no cover
+    from constants import LANGLIST  # type: ignore
     from rules import apply_rules  # type: ignore
+    from utils import levenshtein_dist  # type: ignore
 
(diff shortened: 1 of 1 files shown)


adbar:
Hi @1over137 it makes perfect sense, thanks for the PR!

I want to merge the larger PR #38 before this one so be prepared for a merge conflict. You just need to add the line again in the worst case.

juanjoDiaz:
Hi @1over137 ,

For my own learning, if you don't mind, why is sorting by value better than sorting by key?

Have you calculated the difference in size for the current dictionaries?

1over137:
> Hi @1over137 ,
> 
> For my own learning, if you don't mind, why is sorting by value better than sorting by key?
> 
> Have you calculated the difference in size for the current dictionaries?

I just copied a snippet from a stackoverflow answer, since I didn't know how to sort dicts. I assume it would not make a noticeable difference either way. The value here is the lemma, which should have many more identical occurrences the keys. If you want to check, you can try both with all of the current dictionaries. It would be interesting to know.

adbar:
@1over137 The repository has been restructured, you can now copy your line to the new file (dictionary pickler). That way you'd keep the credit for the change.

juanjoDiaz:
Still, I wouldn't merge a script from StackOverflow without doing some testing to validate that we are actually gaining something.

adbar:
Yes, I would test it beforehand.

adbar:
I just restarted the branch and added the line, I do see a change while sorting by key but sorting by value does seem to help the compression algorithm a bit (though by not -10% as you say).

PR #42, opened 2023-01-24
refactor: split files into modules

juanjoDiaz:
First part from #38 .

Let's see if this makes things easier.

I'll try to add some comments about why I think that classes are the natural next step.


simplemma/lemmatizer.py
```diff
@@ -338,7 +74,7 @@ def _greedy_search(
     i = 0
     while candidate in datadict and (
         len(datadict[candidate]) < len(candidate)
-        and _levenshtein_dist(datadict[candidate], candidate) <= distance
+        and levenshtein_dist(datadict[candidate], candidate) <= distance
```
juanjoDiaz, line 77: How could we configure this without classes and without passing the `levenshtein_dist` method as argument to pretty much every method in this file?

simplemma/langdetect.py
```diff
@@ -4,23 +4,14 @@
...
-def _update_lang_data(lang: Optional[Union[str, Tuple[str]]]) -> None:
-    # convert string
-    lang = _control_lang(lang)
-    # load corresponding data
-    global LANG_DATA
-    if not LANG_DATA or tuple(l.code for l in LANG_DATA) != lang:
-        LANG_DATA = _load_data(lang)
-
-
 def prepare_text(text: str) -> List[str]:
```
juanjoDiaz, line 15: What if we want to offer the users a possibility to customize the sampling logic?
How could we do this without globally overriding this method?

simplemma/lemmatizer.py
```diff
@@ -561,7 +297,7 @@ def lemmatize(
     if not silent:
         raise ValueError(f"Token not found: {token}")
     # try to simply lowercase # and len(token) < 10 ?
-    return token.lower() if lang[0] in BETTER_LOWER else token
+    return token.lower() if dictionaries[0].code in BETTER_LOWER else token
 
 
 def text_lemmatizer(
```
juanjoDiaz, line 303: This method uses the hardcoded `simple_tokenizer` in line 313.
What if the user needs to use a different tokenizer?
Either he has to override the `simple_tokenizer` method or implement a new `text_lemmatizer` using the raw `lemmatizer`

adbar:
Thanks!

simplemma/dictionaries.py
```diff
@@ -0,0 +1,71 @@
...
+            continue
+        LOGGER.debug("loading %s", lang)
+        langlist.append(LangDict(lang, _load_pickle(lang)))
+    return langlist
+
+
+IM_MEMORY_DICTIONARIES: List[LangDict] = []
+
+
+def update_lang_data(lang: Optional[Union[str, Tuple[str]]]) -> List[LangDict]:
```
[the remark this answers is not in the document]
adbar: Maybe we don't need to add caching right here, I avoided it on purpose because passing a whole language dictionary as cache would be a waste of resources.
The current architecture caches single words and language codes, which has a much smaller footprint.

simplemma/langdetect.py
```diff
@@ -43,10 +34,10 @@ def in_target_language(text: str, lang: Optional[Tuple[str]] = None) -> float:
     """Determine which proportion of the text is in the target language(s)."""
     total = 0
     in_target = 0
-    _update_lang_data(lang)
+    dictionaries = update_lang_data(lang)
```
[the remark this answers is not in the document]
adbar: Yes and I believe this point is covered in #33 so we can continue the discussion there. For my purpose I'd say we iterate on the language codes and load/unload the dictionaries. But maybe a more complex caching mechanism would be beneficial (I'm not sure).

simplemma/lemmatizer.py
adbar: Maybe we could split this part from the rest because it's not used by non-greedy searches. The problem with the cache has been addressed and solved in #43. Feel free to open an issue if something is missing.

simplemma/langdetect.py
adbar: I'm not sure what that means, but we can disuss it further in #33 if necessary.

adbar: You're right, I just opened a corresponding issue.

simplemma/lemmatizer.py
adbar: I get your point but I didn't imagine people would actually rely on this function because it's so easy to use any existing tokenizer and then glue it to the lemmatization function. So I don't expect advanced users to work with this function but I'll think about it.

Feel free to open an issue since I understand you may be concerned (i.e. you prefer to use a customized function offered by Simplemma rather than an external one).

simplemma/dictionaries.py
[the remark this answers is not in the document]
juanjoDiaz: This is the main problem I have. Without this, I cannot use this library (at the moment we are using my fork because of this).

We need to control how languages are loaded from disk and kept in memory.

LRU caching is a possibility.
Caching the most called dictionaries is another option.
Preloading certain languages that are used most often and keeping those always in memory is also an option.

All that is not possible to control unless we have some sort of customizable Dictionary Factory.
How do you suggest to move forward?

PR #46, opened 2023-01-26
Add function for language data loading/unloading (#33)

adbar:
Experimental change provided for reference, there could be a better way to solve the problem.


adbar:
Not necessary anymore.

PR #47, opened 2023-01-27
feat: wrap DictionaryFactory in a class

### LINKED ISSUE #33: Add more control on how dictionaries are loaded to memory

juanjoDiaz:
Supersedes #46 and closes #33

At the end, we are simply opting for passing these are additional argument to the functions.


juanjoDiaz:
With this, #33 is not needed since the LRU cache already has the dictionaries cached in-memory.
So, when calling the loading function again it will just take it from there.

simplemma/langdetect.py
```diff
@@ -30,15 +30,21 @@ def prepare_text(text: str, splitting_regex: Pattern[str] = SPLIT_INPUT) -> List
     return [item[0] for item in counter.most_common(1000)]
 
 
-def in_target_language(text: str, lang: Optional[Tuple[str]] = None) -> float:
+def in_target_language(
+    text: str,
+    lang: Optional[Tuple[str]] = None,
+    dictionaryFactory: DictionaryFactory = DictionaryFactory(),
```
adbar, line 36: "Names should be lowercase, with words separated by underscores as necessary to improve readability" as of PEP 8, so _dictionary_factory_ would be more pythonic.

adbar:
Thanks, the PR looks good!

Regarding the contents we could maybe add a function to unload the superfluous dictionaries in case the user asks for a subset of what's already in memory. Say people load a lot of dicts and then wish to load a list larger than cache limit, e.g. `("ast", "de", "en", "es", "fr", "it", "nl", "pt")` → `("es", "fr", "it")`
It would then be beneficial to go through an unloading phase instead of re-loading every language, what do you think?

 I also have a small comment on style (see above).

simplemma/dictionary_factory.py
```diff
@@ -0,0 +1,57 @@
...
+    filename = f"data/{langcode}.plzma"
+    filepath = str(Path(__file__).parent / filename)
+    with lzma.open(filepath, "rb") as filehandle:
+        pickled_dict = pickle.load(filehandle)
+        assert isinstance(pickled_dict, dict)
+        return pickled_dict
+
+
+class DictionaryFactory:
+    def __init__(self, cache_max_size: Optional[int] = 5):
```
adbar, line 36: A power of two for cache size could be better, I suggest 8.
https://stackoverflow.com/questions/61536704/why-does-python-lru-cache-performs-best-when-maxsize-is-a-power-of-two

juanjoDiaz: Solved!

simplemma/langdetect.py
juanjoDiaz: Solved!

juanjoDiaz:
> Regarding the contents we could maybe add a function to unload the superfluous dictionaries in case the user asks for a subset of what's already in memory. Say people want to load a lot of dicts (larger than cache limit), e.g. ("ast", "de", "en", "es", "fr", "it", "nl", "pt") → ("es", "fr", "it")
It would then be beneficial to go through an unloading phase instead of re-loading every language, what do you think?

The problem with that is that it screws up the LRU.

Let's say that your cache is of size 3.
```python
dictionary_factory.get_dictionaries(("en", "de")) // en & de in cache
dictionary_factory.get_dictionaries(("en", "fr")) // en, de & fr in cache (en already in memory)
dictionary_factory.get_dictionaries(("en", "cy")) // de, fr & cy in cache (en already in memory)
dictionary_factory.get_dictionaries(("de", "cy")) // de, fr & cy in cache (en is unloaded)
dictionary_factory.get_dictionaries(("en")) // cy & en in cache (en must be loaded from disk despite having been called so recently)
```

With the simple LRU
```python
dictionary_factory.get_dictionaries(("en", "de")) // en & de in cache
dictionary_factory.get_dictionaries(("en", "fr")) // de, en & fr in cache
dictionary_factory.get_dictionaries(("en", "cy")) // fr, en, & cy in cache
dictionary_factory.get_dictionaries(("de", "cy")) // en, de & cy in cache
dictionary_factory.get_dictionaries(("en")) // en, de & cy in cache (en is already in cache)
```

I've considering implementing a "most often called" cache as an alternative to the LRU cache.

I also have implemented already (didn't include in this PR to keep it smaller) functionality to "preload" a set of languages so they are kept always in memory regardless of the LRU cache rotation.

The options are endless and that's the beauty.
As long as the class implements `def get_dictionaries(self, langs: Optional[Union[str, Tuple[str]]]) -> Dict[str, Dict[str, str]]:`  it will work as a dictionary cache so people can implement whatever strategy they want 🙂

simplemma/dictionary_factory.py
juanjoDiaz: Changed to 4 but changed now to 8

adbar:
Then it's fine for now, thanks for the changes. The PR looks ready, I'll have a look at it again on Monday and eventually merge it.

PR #50, opened 2023-01-31
feat: add posibility to customize token sampler

### LINKED ISSUE #45: Custom token sampling in prepare_text() for language identification
What if we want to offer the users a possibility to customize the sampling logic?
How could we do this without globally overriding this method?

_Originally posted by @juanjoDiaz in https://github.com/adbar/simplemma/pull/42#discussion_r1085860860_

juanjoDiaz:
Closes #45


simplemma/langdetect.py
```diff
@@ -12,34 +12,34 @@
...
+        )
+
+        return [item[0] for item in counter.most_common(1000)]
 
 
 def in_target_language(
     text: str,
     lang: Optional[Tuple[str]] = None,
     dictionary_factory: DictionaryFactory = DictionaryFactory(),
+    tokenSampler: TokenSampler = TokenSampler(),
```
adbar, line 36: This is more JavaScript style, `token_sampler` would be a better variable name here.

PR #52, opened 2023-01-31
refactor: simplify english rules

juanjoDiaz:


adbar:
Thanks, the rules are still pretty experimental but this code is much cleaner.

I compared speed with a regex and it seems the latter is faster (approx. by a factor of 1.5):

```
python3 -m timeit -s 'token = "examinations"; ENGLISH_S_ENDING = ["doms", "isms", "ists", "ments", "nces", "ships", "tions", "ums"]' 'any(token.endswith(e) for e in ENGLISH_S_ENDING); token[:-1]'
```
vs.
```
python3 -m timeit -s 'token = "examinations"; import re; r = re.compile("(dom|ism|ist|ment|nces|ship|tion|um)s")
' 'r.search(token); r.sub("\1", token)'
```

This advantage get larger (factor of 3) exists when using other words including negative matches (just the search part as it's more costly):

```
python3 -m timeit -s 'tokens = ["kingdoms", "kingdom", "hardships", "hardship", "examination", "examinations"]; ENGLISH_S_ENDING = ["doms", "isms", "ists", "ments", "nces", "ships", "tions", "ums"]' 'for t in tokens: any(t.endswith(e) for e in ENGLISH_S_ENDING)'
```
vs.
```
python3 -m timeit -s 'tokens =  ["kingdoms", "kingdom", "hardships", "hardship", "examination", "examinations"]; import re; r = re.compile("(dom|ism|ist|ment|nces|ship|tion|um)s")' 'for t in tokens: r.search(t)'
```

So since we're at it I'd suggest implementing a regex instead. Do you want to do it or should I take care of the modification?

juanjoDiaz:
I can do it.

I tested it and the faster seems to be using a regex for searching but not for substiturion:

python3 -m timeit -s 'token = "examinations"; import re; r = re.compile("(dom|ism|ist|ment|nces|ship|tion|um)s")' 'r.search(token); token[:-1]'

adbar:
That makes sense, let's go for it.

Issue #44, opened 2023-01-25
Make tokenization configurable


adbar:
So far tokenization works with a fixed regular expression coded as a global variable.

To change this behavior if the users wish to do so, we could
- either supply a regex as function argument (the current one being the default regex)
- or wrap the tokenizer function in a configurable Tokenizer class
- _(or do both)_

adbar:
It's now possible to define a regex to split the input. I don't know if it's necessary to go further considering the number of different tokenizers around. I'll leave the thread open for now.

adbar:
@juanjoDiaz If it makes sense to you to customize Simplemma instead of using another software package for tokenization I'm open to review a PR.

PR #53, opened 2023-01-31
refactor: simple_tokenizer to return strings instead of matches

juanjoDiaz:


juanjoDiaz:
>  I wonder why you chose to change the default behavior of simple_tokenizer()? I'd prefer it to stay the same.

This is a pre-step to make the tokenizer configurable just as the `DictionaryFactory` or the `TokenSampler`
I was planning to do a separate PR wrapping the tokenizer in a class and making it configurable.

Conceptually, a tokenizer receives a text (string) and return a list of tokens (strings).
Returning the regex matches is leaking the implementation details of your specific tokenizer. And they are never used. Only the actual text is.

Imagine that I want to use my own tokenizer which is richer than the `simple_tokenizer`. Let's say that is processes hyphens or character normalization and that's not done through regex.
Would I need to wrap my resulting tokens in an array? That doesn't make sense to me.

> After a few tests my impression was that findall is faster than finditer + list, that's why I chose to make findall the default, let the users choose and offer a lemma_iterator() function separately (this refactoring looks good btw).

I guess that the CPU/memory advantages/disadvantages will depend on the size of the text. Iterators are design to minimize memory footprint.
I think that making everything work with iterables is a good appraoch. It's trivial for users to move everything to lists.

Personally, I would just remove `lemma_iterator` and make `text_lemmatizer` return iterator. I didn't change that to keep compatibility.

juanjoDiaz:
Hi,

Reading up about this performance difference between `finditer` and `findall`, it seems that the gains are different depending on the number of matches and the size of the input string.

In any case, the `simple_tokenizer` is hardcoded in both `text_lemmatizer` and `lemma_iterator` to always use `finditer`. So, the potential performance gain from `findall` is never used in practice. That's why I removed it.

I've added the customizable Tokenizer class so you see the full picture.

Wdyt?

juanjoDiaz:
This PR now closes #44

adbar:
I tried different Python versions and different text lengths, the speed difference between `finditer` and `findall` was consistent.

I prefer to keep the `simple_tokenizer()` function as `splitting_regex.findall(text)`. It's faster and that way people already using it don't suddently end up with an iterator instead of a list. Could you please revert this one to its previous state?

The rest looks good, I just have a remark on the tokenizer name.

juanjoDiaz:
> I prefer to keep the simple_tokenizer() function as splitting_regex.findall(text). It's faster and that way people already using it don't suddently end up with an iterator instead of a list. Could you please revert this one to its previous state?

Sure thing. I will do that.
However, That is not how it used to work before.
Before it was always using the `iterate` true. So, it was always using the iterator and never the list approach. 🙂

simplemma/lemmatizer.py
```diff
@@ -319,36 +320,19 @@ def text_lemmatizer(
 ) -> List[str]:
     """Convenience function to lemmatize a text using a simple tokenizer.
     Returns a list of tokens and lemmata."""
-    lemmata = []
-    last = "."  # beginning is initial
-    for match in simple_tokenizer(text, iterate=True):
```
juanjoDiaz, line 324: Here it used the iterator even though it's the list method.

simplemma/lemmatizer.py
```diff
@@ -319,36 +320,19 @@ def text_lemmatizer(
...
     text: str,
     lang: Optional[Union[str, Tuple[str]]] = None,
     greedy: bool = False,
     silent: bool = True,
+    tokenizer: Tokenizer = Tokenizer(),
 ) -> Iterator[str]:
     """Convenience function to lemmatize a text using a simple tokenizer.
     Returns a list of tokens and lemmata."""
-    last = "."  # beginning is initial
-    for match in simple_tokenizer(text, iterate=True):
```
juanjoDiaz, line 348: Here it also used the iterator. So, the list was never used anywhere 😅

adbar:
It will be much better now, before that the situation was confusing.

As you say internally the iterator was used but the default was `iterate=False`.

See here in the Readme: https://github.com/adbar/simplemma#tokenization So far the defaults are:
- `simple_tokenizer`: list
- `text_lemmatizer`: list
- `lemma_iterator`: generator

simplemma/lemmatizer.py
adbar: Within this function the iterator can stay.

simplemma/lemmatizer.py
adbar: Same here, it can stay, I just want the simple_tokenizer function to return a list with `findall`, the other code blocks follow another logic, I'm with you on this.

adbar:
Design decision on my side, we don't have to agree on everything :)
I personally use the `simple_tokenizer()` and I'd like to keep it as fast as it was.

I'll merge this PR in the current state, we can work on another one to make further changes.

PR #54, opened 2023-02-02
refactor: solve linting issues and simplify some methods complexity

juanjoDiaz:
This solves many of the complaints from flake8.

The diff might seems big but if you look closely it's mainly removing nested conditions in favour of guard clauses.

As you see, there are no real changes on the tests. Everything just pass all the same.


adbar:
Thanks, I'll have a look. I use pylint but usually not flake8.

Since we're at it, could you please try to fix this mypy warning? I'm not sure how:
`$ mypy --strict simplemma/`
`simplemma/dictionary_factory.py:47: error: Non-overlapping equality check (left operand type: "Tuple[str, ...]", right operand type: "List[str]")  [comparison-overlap]`

adbar:
I'll also put this pylint refactoring hint on the to-do list (to be tackled by you or me):
`simplemma/dictionary_pickler.py:134:13: R1721: Unnecessary use of a comprehension, use dict(sorted(mydict.items(), key=lambda item: item[1])) instead. (unnecessary-comprehension)`

juanjoDiaz:
fixed!

Also fixed some issues when running mypy on tests.

adbar:
mypy apparently catches errors in the test files.

tests/test_lemmatizer.py
```diff
@@ -10,18 +10,18 @@
 logging.basicConfig(level=logging.DEBUG)
 
 
-def test_custom_dictionary_factory():
+def test_custom_dictionary_factory() -> None:
     class CustomDictionaryFactory:
-        def get_dictionaries(self, langs: Optional[Union[str, Tuple[str]]]):
-            return {"en": {"testing": "the test works!!"}}
+        def get_dictionaries(self, langs: Optional[Union[str, Tuple[str]]]) -> None:
```
adbar, line 15: @juanjoDiaz Could you please look into this?

adbar:
I restored the original expression `token[0].isupper()` (it changed to `token.isupper()` in the process). This could also be tackled directly by the regex.

The example was not functioning anymore so I found another one.

simplemma/dictionary_factory.py
```diff
@@ -33,20 +33,26 @@ def _load_dictionary_from_disk(langcode: str) -> Dict[str, str]:
...
     ) -> Dict[str, Dict[str, str]]:
+        # check validity of input
         langs = _control_lang(langs)
 
-        if self._data and tuple(sorted(self._data.keys())) == sorted(langs):
+        # check if the languages already loaded can be re-used
+        if sorted(self._data.keys()) == sorted(langs):
             return self._data
 
+        # (re)load data
```
juanjoDiaz, line 55: Same thing here.

If you want to make this more clear, we could move this block to an internal method called `load_dictionaries` and get rid of the comment

juanjoDiaz (commented):
Comments simply describing what the code is doing are considered a code smell and a bad practice.

Comments should only be used to document why things are done in a certain way or other aspect not already visible in the code.

simplemma/dictionary_factory.py
```diff
@@ -33,20 +33,26 @@ def _load_dictionary_from_disk(langcode: str) -> Dict[str, str]:
...
         self._load_dictionary_from_disk = lru_cache(maxsize=cache_max_size)(
             _load_dictionary_from_disk
         )
 
     def get_dictionaries(
-        self, langs: Optional[Union[str, Tuple[str]]]
+        self,
+        langs: Optional[Union[str, Tuple[str, ...]]] = None,
     ) -> Dict[str, Dict[str, str]]:
+        # check validity of input
```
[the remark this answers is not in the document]
adbar: Good idea, do you want to do it?

simplemma/dictionary_factory.py
```diff
@@ -33,20 +33,26 @@ def _load_dictionary_from_disk(langcode: str) -> Dict[str, str]:
...
         langs = _control_lang(langs)
 
-        if self._data and tuple(sorted(self._data.keys())) == sorted(langs):
+        # check if the languages already loaded can be re-used
```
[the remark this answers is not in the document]
adbar: OK

simplemma/dictionary_factory.py
adbar: OK, let's remove the comment

adbar:
Never mind, ready to merge.

PR #58, opened 2023-02-03
performance: move rules to dictionary

juanjoDiaz:
Doing a dictionary lookup should be much faster than the if-else block.

The mechanism to see the languages with rules is more solid just looking at the dictionary keys.

I've applied the same logic to the prefix processing.

My next step is to create a `rules` folder where each language has its own file exporting the methods to apply the rules and apply the prefixes.

With all this is much easier to add new rules and manage them.
It's also a first step to wrap rules in a class and make them configurable.

### DIFF
--- a/simplemma/dictionary_pickler.py
+++ b/simplemma/dictionary_pickler.py
@@ -8,12 +8,12 @@ from typing import Dict, List, Optional
 
 try:
     from .constants import LANGLIST
-    from .rules import apply_rules
+    from .rules import APPLY_RULES
     from .utils import levenshtein_dist
 # local error, also ModuleNotFoundError for Python >= 3.6
 except ImportError:  # pragma: no cover
     from constants import LANGLIST  # type: ignore
-    from rules import apply_rules  # type: ignore
+    from rules import APPLY_RULES  # type: ignore
     from utils import levenshtein_dist  # type: ignore
 
 LOGGER = logging.getLogger(__name__)
@@ -77,8 +77,10 @@ def _read_dict(filepath: str, langcode: str, silent: bool) -> Dict[str, str]:
             if len(columns[0]) > 6 and len(columns[1]) == 1:
                 continue
             # tackled by rules
-            if len(columns[1]) > 6:  # columns[1] != columns[0]
-                rule = apply_rules(columns[1], langcode)
+            if (
+                len(columns[1]) > 6 and langcode in APPLY_RULES
+            ):  # columns[1] != columns[0]
+                rule = APPLY_RULES[langcode](columns[1], False)
                 if rule == columns[0]:
                     continue
                 if rule is not None and rule != columns[1]:
--- a/simplemma/lemmatizer.py
+++ b/simplemma/lemmatizer.py
@@ -9,13 +9,8 @@ from typing import Any, Dict, List, Iterator, Optional, Tuple, Union
 from .constants import CACHE_SIZE
 from .dictionary_factory import DictionaryFactory
 from .tokenizer import Tokenizer
+from .rules import APPLY_RULES, FIND_PREFIXES
 from .utils import levenshtein_dist
-
-try:
-    from .rules import apply_rules, GERMAN_PREFIXES, RULES_LANGS, RUSSIAN_PREFIXES
-# local error, also ModuleNotFoundError for Python >= 3.6
-except ImportError:  # pragma: no cover
-    from rules import apply_rules, RULES_LANGS  # type: ignore
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


simplemma/lemmatizer.py
```diff
@@ -217,8 +188,8 @@ def _return_lemma(
     # dictionary search
     candidate = _simple_search(token, datadict, initial=initial)
     # simple rules
-    if candidate is None and lang in RULES_LANGS:
-        candidate = apply_rules(token, lang, greedy)
+    if candidate is None and lang is not None and lang in RULES.keys():
```
adbar, line 191: you can delete `.keys()` here

simplemma/rules.py
```diff
@@ -160,7 +145,7 @@ def apply_de(token: str, greedy: bool = False) -> Optional[str]:
 ENGLISH_S_ENDING = re.compile(r"(dom|ism|ist|ment|nce|ship|tion|um)s$")
 
 
-def apply_en(token: str) -> Optional[str]:
+def apply_en(token: str, greedy: bool = False) -> Optional[str]:
```
adbar, line 148: We could remove `greedy` here and thereafter, it's not used and I don't think it will be

simplemma/rules.py
```diff
@@ -499,3 +488,18 @@ def apply_ru(token: str) -> Optional[str]:
...
             return token[: -len(ending)] + base
     return None
+
+
+PREFIXES: Dict[str, Callable[[str], Optional[str]]] = {
+    "de": apply_prefixes_de,
+    "ru": apply_prefixes_ru,
+}
+
+RULES: Dict[str, Callable[[str, bool], Optional[str]]] = {
```
adbar, line 498: I would prefer a variable name suggesting that it's something dynamic, it isn't just a dictionary. We could keep `APPLY_RULES`, don't you think?

simplemma/rules.py
```diff
@@ -499,3 +488,18 @@ def apply_ru(token: str) -> Optional[str]:
         if token.endswith(ending):
             return token[: -len(ending)] + base
     return None
+
+
+PREFIXES: Dict[str, Callable[[str], Optional[str]]] = {
```
adbar, line 493: Same here, there are functions behind, maybe `FIND_PREFIXES` then?

simplemma/lemmatizer.py
```diff
@@ -233,8 +204,12 @@ def _return_lemma(
     if len(token) <= limit:
         return candidate
     # subword decomposition: predefined prefixes (absent from vocabulary if they are not words)
-    if candidate is None:
-        candidate = _prefix_search(token, lang, datadict)  # type: ignore[arg-type]
+    if candidate is None and lang in PREFIXES:
+        prefix = PREFIXES[lang](token)
```
adbar, line 208: These lines could be listed in a separate function since we're re-organizing the code

adbar:
Yes, a dictionary is supposed to be more efficient as explained [here](https://stackoverflow.com/questions/15923766/python-dictionary-vs-if-statement-speed).

I don't see a speed difference right now but it makes it easier to organize the code, I'd say we can go on with this.

simplemma/rules.py
juanjoDiaz: The reason to have is so all the method have the same typings in the dictionary.
Otherwise, there are errors.

simplemma/lemmatizer.py
```diff
@@ -9,14 +9,9 @@
 from .constants import CACHE_SIZE
 from .dictionary_factory import DictionaryFactory
 from .tokenizer import Tokenizer
+from .rules import FIND_PREFIXES, APPLY_RULES
```
adbar, line 12: small detail: the imports aren't alphabetically sorted

tests/test_rules.py
```diff
@@ -2,11 +2,23 @@
 
 import logging
 
-from simplemma.rules import apply_rules, apply_de, apply_en, apply_fi, apply_nl
+from simplemma.rules import (
```
adbar, line 5: alphabetical sort here too

adbar:
You created an empty file `rules/de.py`, it the PR ready for a merge or are you still working on it?

simplemma/dictionary_pickler.py
```diff
@@ -77,8 +77,10 @@ def _read_dict(filepath: str, langcode: str, silent: bool) -> Dict[str, str]:
             if len(columns[0]) > 6 and len(columns[1]) == 1:
                 continue
             # tackled by rules
-            if len(columns[1]) > 6:  # columns[1] != columns[0]
-                rule = apply_rules(columns[1], langcode)
+            if (
+                len(columns[1]) > 6 and langcode in APPLY_RULES.keys()
```
adbar, line 81: You can delete `.keys()`.

juanjoDiaz:
Should be good to merge now.
That de.py file would be the next step.

adbar:
Nice, thanks!

PR #59, opened 2023-02-03
refactor: rename langdetect to language_detector to be more clear and…

juanjoDiaz:
… better match the other modules in the package


osma:
I hope you realized that this is an API change from the library user point of view. For example Annif [imports from simplemma.langdetect](https://github.com/NatLibFi/Annif/blob/8a194c4b0bb8bf198ff4fb35a4fb6fa1f367cd3f/annif/transform/langfilter.py#L4) as was previously suggested in the README and this will break when we upgrade to the next Simplemma release that includes this change.

This is no big deal for us, we will of course adjust the import. Simplemma is still in the 0.x version range when changes like this are permitted. But it would be good to mention API changes like this in the release notes.

adbar:
Hi @osma, thanks for the hint, I'm planning to keep everything running as it was. I'll keep track of changes (see issue above) and add deprecation warnings.

PR #60, opened 2023-02-06
More opportunistic language detection

adbar:


simplemma/language_detector.py
```diff
@@ -22,10 +22,13 @@ def __init__(self, tokenizer: Tokenizer = Tokenizer(SPLIT_INPUT)) -> None:
...
         some of the rest, and return at most 1000 tokens."""
-        # generator expression to split the text
-        counter = Counter(
-            token for token in self.tokenizer.split_text(text) if not token[0].isupper()
-        )
+
+        counter = Counter(token for token in self.tokenizer.split_text(text))
+
+        deletions = [token for token in counter if token[0].isupper()]
+        if len(deletions) < 0.8 * len(counter):
```
juanjoDiaz, line 29: I think that this should be configurable by the user (through the constructor) and documented in the docs :)

adbar:
Ready to merge or do you see something else @juanjoDiaz?

simplemma/language_detector.py
```diff
@@ -10,24 +10,44 @@
...
-        # generator expression to split the text
-        counter = Counter(
-            token for token in self.tokenizer.split_text(text) if not token[0].isupper()
-        )
 
-        return [item[0] for item in counter.most_common(1000)]
+        counter = Counter(token for token in self.tokenizer.split_text(text))
+
+        deletions = [token for token in counter if token[0].isupper()]
+        if len(deletions) < self.capitalized_threshold * len(counter):
```
juanjoDiaz, line 38: This can delete up to 80% of the text.
Is that what you wanted? Seems like a lot.

Btw, you could also  sample here the max between deletions & capitalized_threshold and delete that, so you can delete something even if not everything.

simplemma/language_detector.py
```diff
@@ -10,24 +10,44 @@
...
         some of the rest, and return at most 1000 tokens."""
-        # generator expression to split the text
-        counter = Counter(
-            token for token in self.tokenizer.split_text(text) if not token[0].isupper()
-        )
 
-        return [item[0] for item in counter.most_common(1000)]
+        counter = Counter(token for token in self.tokenizer.split_text(text))
+
+        deletions = [token for token in counter if token[0].isupper()]
```
juanjoDiaz, line 37: This whole block could be wrapped in `if self.capitalized_threshold == 0` so the whole process is skipped if the threshold is 0.

simplemma/language_detector.py
```diff
@@ -10,24 +10,44 @@
...
 
 class TokenSampler:
-    __slots__ = ["tokenizer"]
+    __slots__ = ["capitalized_threshold", "max_tokens", "tokenizer"]
 
-    def __init__(self, tokenizer: Tokenizer = Tokenizer(SPLIT_INPUT)) -> None:
+    def __init__(
+        self,
+        tokenizer: Tokenizer = Tokenizer(SPLIT_INPUT),
+        max_tokens: int = 1000,
```
juanjoDiaz, line 24: I'd make `max_tokens` optional. If none, I'd return the full list.

Likewise, I'd make `capitalized_threshold` optional and fi none, I'd removed all capitalized words.

simplemma/language_detector.py
juanjoDiaz, line 37: Why do we delete capitalized words?
Aren't we losing the first word of every sentence?

simplemma/language_detector.py
```diff
@@ -10,24 +10,44 @@
...
+
+        deletions = [token for token in counter if token[0].isupper()]
+        if len(deletions) < self.capitalized_threshold * len(counter):
+            for token in deletions:
+                del counter[token]
+
+        return [item[0] for item in counter.most_common(self.max_tokens)]
+
+
+class RelaxedTokenSampler(TokenSampler):
```
juanjoDiaz, line 45: With what I proposed above, this could be done just as
```python
class RelaxedTokenSampler(TokenSampler):
    def __init__(self) -> None:
        super().__init__(RELAXED_SPLIT_INPUT, None, None)
```
right?


Or just during constructor

```python
TokenSampler(RELAXED_SPLIT_INPUT, None, None)
```

Instead of creating a bunch of TokenSampler we could have the basic one and just document different possible configurations.
See my comment below.

simplemma/language_detector.py
```diff
@@ -91,10 +111,13 @@ def lang_detector(
...
-        if len(results) > 1 and results[0][1] == results[1][1]:
-            return _return_default()
+        # switch unknown first item to the end
+        if results[0][0] == "unk":
+            pair = results.pop(0)
+            results.append(pair)
+        # in case of ex-aequo use other token sampling to discriminate
+        if not greedy and results[0][1] == results[1][1]:
+            results = lang_detector(
+                text, lang=lang, greedy=True, token_sampler=RelaxedTokenSampler()
```
juanjoDiaz, line 121: I'm a bit conflicted here.
We let the user provide a `TokenSampler`. But then there is a case in which we hardcode a separate TokenSampler.

Hardcoded things remove flexibility in general.

I would say that we should return whatever the TokenSampler produces even if there is a tie.
We could document, that in case of a tie, you can try again a different TokenSampler (and document how to configure the RelaxedTokenSampler as I suggested above)

simplemma/language_detector.py
```diff
@@ -91,10 +111,13 @@ def lang_detector(
...
     # post-processing
     if len(results) > 1:
-        # in case of ex-aequo
-        if greedy is False and results[0][1] == results[1][1]:
-            results = lang_detector(text, lang=lang, greedy=True)
-        # fallback
-        if len(results) > 1 and results[0][1] == results[1][1]:
-            return _return_default()
+        # switch unknown first item to the end
+        if results[0][0] == "unk":
```
juanjoDiaz, line 115: I think that it's good to keep this sorted.
What if the unknown result is the second and not the first?

You can simply check if the first or second result is unk. And skip that.

simplemma/language_detector.py
```diff
@@ -91,10 +111,13 @@ def lang_detector(
...
+            pair = results.pop(0)
+            results.append(pair)
+        # in case of ex-aequo use other token sampling to discriminate
+        if not greedy and results[0][1] == results[1][1]:
```
juanjoDiaz, line 119: What if there is only a single result?
The length check is relevant.

juanjoDiaz:
> on my data the language detection is then faster and more accurate.

Not directly related to this PR, I think that at some point there should be some documentation about how are the dictionaries created and validated, where are the rules coming from, how are you testing/evaluating all this, etc.
So it's easier to understand why things are the way they are

simplemma/language_detector.py
adbar: Sampling between both operations can be detrimental in case more words are needed in the end, and it may also be slower. I'm deciding against it for now.

adbar:
Concerning the default settings, I'm trying to optimize the detector on my benchmark ([short texts extracted from Wikipedia](https://zenodo.org/record/841984), a good test case IMO) so that users who don't want to customize everything don't have to. That's what leads to certain design decisions above.

The recent improvements have to do with recall: don't delete capitalized words if it means deleting almost everything, return something instead of "unk" (even if it's the first element), trying to disambiguate between two languages in case of ex-aequo.

Yes, the docs could be extended.

simplemma/language_detector.py
adbar: Or just put the unknown proportion systematically at the end?

simplemma/language_detector.py
adbar: I'm more inclined to further reduce the value for max_tokens to 100. It's an important feature, default detection will be too slow on long texts otherwise.

adbar:
@osma you use the language detector concerned by this PR. Apart from the function name it's functionality for you should stay untouched, it's mostly getting better behind the scenes.

Since we're reviding part of its logic, do you have anything to share? I'm thinking of classification mistakes you noticed or settings you'd like to be able to configure.

osma:
@adbar Thanks for asking. We're mainly using Simplemma language detection as a relatively cheap heuristic in determining whether a sentence is in the expected language or not. A typical use case is text extracted from a PDF thesis that contains abstracts in multiple languages. For example many doctoral theses in Finland are written mainly in English because of demands from an international academic audience, but they must have abstracts in Finnish language as well due to university rules. So we want to skip the abstracts that are in the "wrong" language before feeding the rest of the text to a ML classifier. It's not a catastrophe if the language detection algorithm sometimes makes mistakes, because usually there will still be plenty of text left for the classifier.

What you're doing here seems sensible and I wouldn't expect any bad outcomes from our perspective. Faster and more accurate sound like good improvements to me :)

juanjoDiaz:
Good conversation as always, gents!

So, here is my issues with the current status of Lang detector:
`in_target_language` and `lang_detector` work differently.

`in_target_language` is missing the `greedy option`.
`lang_detector` counts incorrectly the number of unknowns. It counts the max unknown tokens for each language but it doesn't count  the tokens that are actually unknown for all the given languages.

Let's take an example
`in_target_language("opera post physica posita (τὰ μετὰ τὰ φυσικά)", lang="la")` returns 50% which is correct.
However,
`lang_detector("opera post physica posita (τὰ μετὰ τὰ φυσικά)", lang="la")` returns [("unk", 1)] which is not correct imo.

The reality is that the text is half latin, half Greek. And there is no reason to disambiguate.

Let me do a PR to show my thinking.

adbar:
Thanks for your comment, I was also planning to work on `in_target_language()`. I'm going to merge this PR first and then address your PR.

PR #62, opened 2023-02-07
refactor: rename extensive to greedy

juanjoDiaz:
To be consistent.

I'm still not 100% clear what greedy does, though.
Some docs are needed. As requested in an existing issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,6 +81,9 @@ Simplemma is used by selecting a language of interest and then applying the data
     ['hier', 'sein', 'Vaccines']
 
 
+Chaining languages
+~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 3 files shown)


adbar:
I already updated the readme but I'll add something a line or two on language detection here, it's still missing as you say.

adbar:
I'd like to push the readme update but cannot push to your repository anymore, is it normal?

juanjoDiaz:
I normally don't allow maintainers to push to me repos (it could lead to password leakage if I put any password in my fork. Not the case).

I gave you permissions now 🙂

adbar:
Strange, I pushed on your branches before, but I understand, thanks for the change.

Is the readme clearer now?

juanjoDiaz:
Yup. All good!

Issue #51, opened 2023-01-31
Use require.txt for dependencies and dev dependencies


juanjoDiaz:
I think that this is a must to ensure that other people can easily get the library on their local.

At the moment there is no prod dependencies so that's not a concern.

osma:
Isn't it enough to have the dependencies declared in `setup.py` ? Simplemma is a library, not an end user application. Tools like setuptools and pip are able to pick up those dependencies from setup.py.

Of course it would be possible to switch to a smarter dependency management tool such as Poetry, and/or a more modern declaration of dependencies like `pyproject.toml` (PEP 518, supported by e.g. Poetry).

adbar:
The good thing is that there are no required dependencies.

The requirements for tests and evaluation could be supplied in a separate file, in the future I would support a move towards PEP 518 if this solution becomes prevalent.

PR #65, opened 2023-02-09
refactor: language detector improvements

juanjoDiaz:
Related to #60

This PR does the following:
* Wraps the `in_target_language` and `lang_detector` methods in a class
* Fix the `lang_detector` method in the class so it count unknown tokens correctly.
* Refactor the `in_target_language` method in the class to use the `lang_detector` (perfecto consistency between both results)

The second attempts to disambiguate and any other attempt that you can do have been left within the "legacy" `lang_detector` method.
Here you could also introduce your attempt with the different Sampler.

So, a user could use the class, and could do as many disambiguations as he wants while maintaining control of the process.
This class could be extended to receive a list of samplers to try in order in case of ex-aequo. Or a setting to try greedy if non-greedy search produced a tie. Or any other advance configuration.

Does this make sense?

### DIFF
--- a/simplemma/__init__.py
+++ b/simplemma/__init__.py
@@ -7,7 +7,7 @@ __license__ = "MIT"
 __version__ = "0.9.1"
 
 
-from .language_detector import in_target_language, lang_detector
+from .language_detector import in_target_language, langdetect
 from .lemmatizer import lemmatize, lemma_iterator, text_lemmatizer, is_known
 from .tokenizer import simple_tokenizer
 from .token_sampler import TokenSampler
--- a/simplemma/__init__.py
+++ b/simplemma/__init__.py
@@ -10,5 +10,6 @@ __version__ = "0.9.1"
 from .language_detector import in_target_language, lang_detector
 from .lemmatizer import lemmatize, lemma_iterator, text_lemmatizer, is_known
 from .tokenizer import simple_tokenizer
+from .token_sampler import TokenSampler
 from .dictionary_factory import DictionaryFactory
 from .dictionary_pickler import *
--- a/simplemma/language_detector.py
+++ b/simplemma/language_detector.py
@@ -62,7 +62,13 @@ def _as_list(results: Dict[str, float]) -> List[Tuple[str, float]]:
 
 
 class LanguageDetector:
-    __slots__ = ["dictionary_factory", "greedy", "lang", "token_sampler"]
+    __slots__ = [
+        "dictionary_factory",
+        "greedy",
+        "lang",
+        "_orig_token_sampler",
+        "token_sampler",
+    ]
 
     def __init__(
         self,
@@ -75,6 +81,10 @@ class LanguageDetector:
         self.greedy = greedy
         self.dictionary_factory = dictionary_factory
         self.token_sampler = token_sampler
+        self._orig_token_sampler = token_sampler
+
+    def _restore_token_sampler(self) -> None:
+        self.token_sampler = self._orig_token_sampler
 
     def proportion_in_each_language(
         self,
@@ -125,15 +135,18 @@ class LanguageDetector:
     def main_language(
         self,
         text: str,
-        token_samplers: List[TokenSampler] = [
-            MostCommonTokenSampler(),
-            RelaxedMostCommonTokenSampler(),
+        additional_token_samplers: List[TokenSampler] = [
+            RelaxedMostCommonTokenSampler()
         ],
     ) -> str:
+        token_samplers = [self.token_sampler] + additional_token_samplers
+
         for token_sampler in token_samplers:
             self.token_sampler = token_sampler
             list_results = _as_list(self.proportion_in_each_language(text))
             if len(list_results) > 1 and list_results[0][1] != list_results[1][1]:
+                self._restore_token_sampler()
                 return list_results[0][0]
 
+        self._restore_token_sampler()
         return "unk"
--- a/simplemma/language_detector.py
+++ b/simplemma/language_detector.py
@@ -86,24 +86,23 @@ def lang_detector(
     results = LanguageDetector(
         dictionary_factory, token_sampler
     ).get_text_percentage_in_each_language(text, lang, greedy)
-    # post-processing
-    if len(results) == 1:
-        return _convert_results_to_sorted_list(results)
-    # in case of ex-aequo
     list_results = _convert_results_to_sorted_list(results)
+
+    # post-processing
+    if len(list_results) == 1:
+        return list_results
+    # in case of ex-aequo
     if greedy is False and list_results[0][1] == list_results[1][1]:
         results = LanguageDetector(
             dictionary_factory, token_sampler
         ).get_text_percentage_in_each_language(text, lang, greedy=True)
-
-    list_results = _convert_results_to_sorted_list(results)
+        list_results = _convert_results_to_sorted_list(results)
     # in case of ex-aequo use other token sampling to discriminate
     if not greedy and list_results[0][1] == list_results[1][1]:
         results = LanguageDetector(
             dictionary_factory, backup_sampler
         ).get_text_percentage_in_each_language(text, lang, greedy=True)
-
-    list_results = _convert_results_to_sorted_list(results)
+        list_results = _convert_results_to_sorted_list(results)
     return list_results
 
 
--- a/simplemma/language_detector.py
+++ b/simplemma/language_detector.py
@@ -4,7 +4,7 @@ import re
 
 from collections import Counter
 from operator import itemgetter
-from typing import List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from .lemmatizer import _return_lemma
 from .dictionary_factory import DictionaryFactory
@@ -55,30 +55,28 @@ class RelaxedTokenSampler(TokenSampler):
 def in_target_language(
     text: str,
     lang: Optional[Union[str, Tuple[str, ...]]] = None,
+    greedy: bool = False,
     dictionary_factory: DictionaryFactory = DictionaryFactory(),
     token_sampler: TokenSampler = TokenSampler(),
 ) -> float:
     """Determine which proportion of the text is in the target language(s)."""
-    total = 0
-    in_target = 0
-    dictionaries = dictionary_factory.get_dictionaries(lang)
-    for token in token_sampler.sample_tokens(text):
-        total += 1
-        for lang_code, lang_dictionary in dictionaries.items():
-            candidate = _return_lemma(
-                token, lang_dictionary, greedy=True, lang=lang_code
-            )
-            if candidate is not None:
-                in_target += 1
-                break
-    if total > 0 and in_target > 0:
-        return in_target / total
-    return 0
+    return LanguageDetector(dictionary_factory, token_sampler).in_target_language(
+        text, lang, greedy
+    )
 
 
-def _return_default() -> List[Tuple[str, float]]:
-    # todo: None if 'unk'?
-    return [("unk", 1)]
+def _convert_results_to_sorted_list(
+    results: Dict[str, float]
+) -> List[Tuple[str, float]]:
+    list_results: List[Tuple[str, float]] = sorted(
+        results.items(), key=itemgetter(1), reverse=True
+    )
+    # switch unknown to the end
+    for i, item in enumerate(list_results):
+        if item[0] == "unk":
+            pair = list_results.pop(i)
+            list_results.append(pair)
+    return list_results
 
 
 def lang_detector(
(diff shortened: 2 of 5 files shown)


adbar:
I'm aware this causes a conflict but I wanted to merge the previous PR to work on one PR at a time on this topic.

It makes sense, could you please adapt it to the current state so we can see if and how continuity of results can be ensured?

juanjoDiaz:
Rebased done.
Not sure why formatting is failing in some of the black linting. It seems fine in my local.

I still don't understand the problem with ex-aequo.
If a sentence is half English & half German, why would not accept that and try to break the tie?
Could you elaborate on this?

Basically, the "legacy" "lang_detector" method tries multiple configurations one after the other.
What's the point of letting the user select the `greedy` option if we are trying both in any case?

adbar:
You may need to update your black version?

In classification tasks you often have to return one label only (other tasks are then called multi-class). If we cannot break the tie then we just won't but it's beneficial to try something else if there is a draw.

I'm not sure about the greedy option either, it doesn't bring much compared to another token sampling strategy.

juanjoDiaz:
That was it. This will be solved by #51 & #54

adbar:
Nice, I'll have a look at it tomorrow.

juanjoDiaz:
> In classification tasks you often have to return one label only (other tasks are then called multi-class). If we cannot break the tie then we just won't but it's beneficial to try something else if there is a draw.

I understand your point but I think that you are mixing 2 problem here.
I would expect a language detector to be a multi-class (or multi-label) classifier and detect the languages based on my settings.
If those languages are 50/50 en & de or 30/30/30 en, de & fr, that's what it is and what should be returned.

If you would like to detect a single language, I would have a separa method that returns a single language code (as str).
In that method, we could do a more aggressive discrimination to get a single language.

I think that currently, my class method is a multi-label language detector and the old method is a single label classifier that returns multiple labels (a bit confusing).

Those this make sense?

simplemma/language_detector.py
```diff
@@ -87,42 +76,102 @@ def lang_detector(
...
+        Perform a first run and further discriminate between the results if necessary.
+        """
+
+        tokens = self.token_sampler.sample_tokens(text)
+        total_tokens = len(tokens)
+        if total_tokens == 0:
+            return {"unk": 1}
+        # iterate
+        dictionaries = self.dictionary_factory.get_dictionaries(lang)
+        known_tokens_count = dict.fromkeys(dictionaries, 0)
```
adbar, line 137: `defaultdict(int)` could be clearer here

simplemma/language_detector.py
```diff
@@ -87,42 +76,102 @@ def lang_detector(
...
+
+    list_results = _convert_results_to_sorted_list(results)
+    # in case of ex-aequo use other token sampling to discriminate
+    if not greedy and list_results[0][1] == list_results[1][1]:
+        results = LanguageDetector(
+            dictionary_factory, backup_sampler
+        ).get_text_percentage_in_each_language(text, lang, greedy=True)
+
+    list_results = _convert_results_to_sorted_list(results)
+    return list_results
```
adbar, line 107: or directly `return _convert_results_to_sorted_list(results)`

simplemma/language_detector.py
```diff
@@ -87,42 +76,102 @@ def lang_detector(
...
-        # in case of ex-aequo use other token sampling to discriminate
-        if not greedy and results[0][1] == results[1][1]:
-            results = lang_detector(
-                text, lang=lang, greedy=True, token_sampler=backup_sampler
-            )
-    return results
+    if len(results) == 1:
+        return _convert_results_to_sorted_list(results)
+    # in case of ex-aequo
+    list_results = _convert_results_to_sorted_list(results)
```
adbar, line 93: In some cases that's three conversions in total, maybe there's a better way.

simplemma/language_detector.py
```diff
@@ -87,42 +76,102 @@ def lang_detector(
...
+
+    def __init__(
+        self,
+        dictionary_factory: DictionaryFactory = DictionaryFactory(),
+        token_sampler: TokenSampler = TokenSampler(),
+    ) -> None:
+        self.dictionary_factory = dictionary_factory
+        self.token_sampler = token_sampler
+
+    def get_text_percentage_in_each_language(
```
adbar, line 121: The name is a bit long but at least it's clear.

adbar:
Thanks, the PR makes sense. I just wonder if it could be worth performing the conversion the other way round: keeping the result list as it was and adding a parameter like `as_dict` if someone wants to see the function return a dict.

I agree that the current situation is confusing as it tries to address single-class and multi-class labels at the same time. 

Third option: since we're revamping it we could as well create a `DetectionResults` class with the desired attributes (e.g. languages loaded, proportion per language and proportion of unknown words). The conversions would be performed directly by a function within the class, like `results.as_dict()`, `results.as_list()`, or `results.best()`. It could be a transparent way to ask for all the results or just for the pair `(best_lang, score)`.

What do you think?

simplemma/language_detector.py
juanjoDiaz: How is `defaultdict` any better than initializing a normal `dictionary` for all the languages starting at `0`?

simplemma/language_detector.py
juanjoDiaz: I've reduced the conversions to a minimum

juanjoDiaz:
> Thanks, the PR makes sense. I just wonder if it could be worth performing the conversion the other way round: keeping the result list as it was and adding a parameter like `as_dict` if someone wants to see the function return a dict.
> 
> I agree that the current situation is confusing as it tries to address single-class and multi-class labels at the same time.
> 
> Third option: since we're revamping it we could as well create a `DetectionResults` class with the desired attributes (e.g. languages loaded, proportion per language and proportion of unknown words). The conversions would be performed directly by a function within the class, like `results.as_dict()`, `results.as_list()`, or `results.best()`. It could be a transparent way to ask for all the results or just for the pair `(best_lang, score)`.
> 
> What do you think?

I think that the natural result of the multi-class classifier is a `dictionary[language, percentage]`.
A sorted list of pairs is just a way of sorting the dictionary (since sorted dictionaries are not a thing in Python).
And something like `results.best()` is just a simple `max(results, key=results.dict)`. 

A `DetectionResults` class doesn't solve the ex-aequo situation when trying to find a single-class classifier. Which is something that you currently solve with multiple runs of the multi-class classified with different configurations


I can think of 2 options:

**Return "unk" when there is a tie and leave the responsibility of trying different TokenSamplers or different greedy settings by documenting it**

```python
    def get_text_language(
        self,
        text: str,
        lang: Optional[Union[str, Tuple[str, ...]]] = None,
        greedy: bool = False,
    ) -> str:
        results = self.get_text_percentage_in_each_language(text, lang, greedy)
        list_results = _convert_results_to_sorted_list(results)
        if len(list_results) > 1 and list_results[0][1] == list_results[1][1]:
            return "unk"
        return list_results[0]
```

The user then can do:

```python
result = LanguageDetector().get_text_language(text)
if result != "unk":
    return result
result =  LanguageDetector(
            backup_sampler = TokenSampler(
                tokenizer=Tokenizer(RELAXED_SPLIT_INPUT),
                max_tokens=1000,
                capitalized_threshold=0,
            )
        ).get_text_percentage_in_each_language(text, lang, greedy=True)

// etc...
```

**Pass the TokenSampler to the methods and not to the constructor of the class**

That way, the `get_text_language` could take a list of TokenSampler and try them in order

```python
    def get_text_language(
        self,
        text: str,
        lang: Optional[Union[str, Tuple[str, ...]]] = None,
        greedy: bool = False,
        token_samplers: List[TokenSampler]
    ) -> str:
        for token_sampler in token_samplers:
            results = self.get_text_percentage_in_each_language(text, lang, greedy, token_sampler)
            list_results = _convert_results_to_sorted_list(results)
            if len(list_results) == 1 or list_results[0][1] != list_results[1][1]:
                return list_results[0]
        return "unk"
```

This would only try multiple samplers and won't try with `greedy` set to `True`.

What do you think?

simplemma/language_detector.py
adbar: It's restricted to ints but it doesn't change much, we can leave it as it is now.

adbar:
Even if `.best()` is just `max(results, key=results.dict)` it could still be worth offering this option.

I'd find option 2 better: use a cascade of different sampling methods, from quick to more refined ones.

simplemma/language_detector.py
```diff
@@ -1,128 +1,179 @@
...
+def langdetect(
     text: str,
     lang: Optional[Union[str, Tuple[str, ...]]] = None,
     greedy: bool = False,
     dictionary_factory: DictionaryFactory = DictionaryFactory(),
-    token_sampler: TokenSampler = TokenSampler(),
-    backup_sampler: TokenSampler = RelaxedTokenSampler(),
+    token_samplers: List[TokenSampler] = [
+        TokenSampler(),
+        TokenSampler(
```
adbar, line 46: This could still be defined above.

simplemma/language_detector.py
```diff
@@ -1,128 +1,179 @@
...
+    )
 
 
-def _return_default() -> List[Tuple[str, float]]:
-    # todo: None if 'unk'?
-    return [("unk", 1)]
-
-
-def lang_detector(
+def _convert_results_to_sorted_list(
```
adbar, line 25: Why not an .as_sorted_list() method since we now have a `LanguageDetector` class?

simplemma/__init__.py
```diff
@@ -7,8 +7,9 @@
 __version__ = "0.9.1"
 
 
-from .language_detector import in_target_language, lang_detector
+from .language_detector import in_target_language, langdetect
```
adbar, line 10: The new name can stay but why not.

simplemma/__init__.py
```diff
@@ -7,8 +7,9 @@
...
 from .lemmatizer import lemmatize, lemma_iterator, text_lemmatizer, is_known
 from .tokenizer import simple_tokenizer
+from .token_sampler import TokenSampler
```
adbar, line 13: I'm not sure why token sampling should be performed by a separate module, it's only used for language detection isn't it?

simplemma/language_detector.py
```diff
@@ -1,128 +1,179 @@
...
+    """Determine which proportion of the text is in the target language(s)."""
+    for token_sampler in token_samplers:
+        results = LanguageDetector(
+            dictionary_factory
+        ).detect_text_proportion_in_each_language(text, lang, greedy, token_sampler)
+        list_results = _convert_results_to_sorted_list(results)
+
+        # post-processing
+        if len(list_results) == 1 or list_results[0][1] != list_results[1][1]:
+            return list_results
```
adbar, line 62: I think `break` would be better here from a syntax perspective.

simplemma/__init__.py
juanjoDiaz: As you want, I thought that for new names, users can use the class and for old names, they can use the old methods.

adbar:
I'm not convinced by your addition of "tokens_proportion" functions, there is not need for that IMO since we give the possibility to customize token sampling. Overriding this should be enough for advanced users. What does this bring?

The function names are unclear. Here are suggestions to shorten function names in `LanguageDetector` (the "detect" is already in the name of the class):
- `detect_text_proportion_in_each_language` →  `proportion_as_dict()` or even `as_dict()`
- `langdetect` could be a function within the class, say `.as_list()`
- `detect_text_proportion_in_languages`: `in_target()` (`in_target_language` was also the original name if I understand correctly)
- `detect_text_main_language` → `.best()`

simplemma/__init__.py
juanjoDiaz: Yup. But it's a separate class.

Users can extend TokenSampler and provide their own sampling mechanisms.
Also, in the future, this library could add other functionality using sampling.

I just tried to keep the 1 file = 1 class approach.

simplemma/language_detector.py
juanjoDiaz: I personally like early returns just as I like guard clauses.

Things like break are more likely to introduce issues if new code is added after the for loop.

But I understand that this is a matter of personal preference. So, I can switch to `break` if you prefer that.

adbar: I think this code shouldn't move much so we could use break.

simplemma/__init__.py
adbar: OK.

simplemma/__init__.py
adbar: OK, let's leave it that way.

simplemma/language_detector.py
adbar: Let's adress the concerns below regarding function names and see afterwards.

juanjoDiaz:
Btw, I'm still not convinced about the further discrimination in case of ex-aequo.

The text `"Dieser Satz ist auf Deutsch. Y esta está en Español."` is 50/50 German & Spanish.
Using the further discrimination with the relaxed token sampler. Yields German `[('de', 0.625), ('es', 0.375), ('unk', 0.0)]` which seems clearly incorrect to me.

I think that the `main_language` method should simply return a `"unk"` if there is a tie.
And it should be up to the user if he wants to do something else.

simplemma/language_detector.py
```diff
@@ -1,128 +1,147 @@
...
+        results["unk"] = unknown_tokens_count / total_tokens
+        return results
+
+    def proportion_in_target_languages(
+        self,
+        text: str,
+        token_sampler: TokenSampler = TokenSampler(),
+    ) -> float:
+        return sum(
+            [
```
adbar, line 119: I'm not sure this line is necessary, does `sum(percentage...` work?

simplemma/language_detector.py
adbar: Could you please address this, i.e. define a relaxed/backup sampler outside of the function?

simplemma/language_detector.py
adbar: Could you please make the change?

adbar:
The function names are clearer, thanks. We can now address the remaining questions one by one.

For performance purposes in benchmark it's better not to return "unknown" even if it would be logical. In this case the backup method is a workaround, your example makes sense but most real-world cases can be decided by using a few tricks. I'd say it's an empirical way of thinking but we can discuss the issue later if problems arise.

simplemma/language_detector.py
adbar: I read a bit more about this (early exit vs break), we can keep it that way.

adbar:
I addressed the remaining concerns above and simplified the code, everything OK?

You introduced a major issue with [typing.Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol), it's only supported by Python 3.8+. We need to find a way around it for earlier versions or replace this code by something else (I'm not sure the NotImplemented statements are needed).

simplemma/token_sampler.py
```diff
@@ -0,0 +1,76 @@
...
+            deletions = [token for token in counter if token[0].isupper()]
+            if len(deletions) < self.capitalized_threshold * len(counter):
+                for token in deletions:
+                    del counter[token]
+
+        return [item[0] for item in counter.most_common(self.sample_size)]
+
+
+class RelaxedMostCommonTokenSampler(MostCommonTokenSampler):
+    __slots__ = ["capitalized_threshold", "sample_size"]
```
juanjoDiaz, line 68: This slots are already inherited from the parent class. So there shouldn+t be a need to have them here.

simplemma/language_detector.py
```diff
@@ -1,128 +1,139 @@
...
+                percentage,
+            ) in self.proportion_in_each_language(text).items()
+            if lang_code != "unk"
+        )
+
+    def main_language(
+        self,
+        text: str,
+        token_samplers: List[TokenSampler] = [
+            MostCommonTokenSampler(),
```
juanjoDiaz, line 129: This was better before.

The class has a token sampler set at the constructor.
And this specific method can provide additional samplers for discrimination.

With this change the, main tokenizer is not used here which is odd.

juanjoDiaz:
> I addressed the remaining concerns above and simplified the code, everything OK?

I've added a couple of comments.
The only thing that I'm not really happy is the change on the `main_language` function.


> You introduced a major issue with [typing.Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol), it's only supported by Python 3.8+. We need to find a way around it for earlier versions or replace this code by something else (I'm not sure the NotImplemented statements are needed).

What are the target Python versions that you aim to support?

The protocol class is just to make the static type-checker more solid.
We can live without it.

simplemma/language_detector.py
adbar: But you were resetting the class sampler which seems cumbersome, maybe there is a better way to do this?

adbar:
I try to support Python 3.6+, 3.7+ is a minimum, there are still a fair amount of people using these versions. So I'd prefer to remove `Protocol`.

simplemma/token_sampler.py
adbar: That's correct, I missed it.

simplemma/language_detector.py
juanjoDiaz:
```python
    def main_language(
        self,
        text: str,
        additional_token_samplers: List[TokenSampler] = [
            RelaxedMostCommonTokenSampler()
        ],
    ) -> str:
        original_token_sampler = self.token_sampler

        for token_sampler in [self.token_sampler] + additional_token_samplers:
            self.token_sampler = token_sampler
            list_results = _as_list(self.proportion_in_each_language(text))
            if len(list_results) > 1 and list_results[0][1] != list_results[1][1]:
                return list_results[0][0]

        self.token_sampler = original_token_sampler

        return "unk"
```

The logic here is that there is `self.token_sampler` which is used by all methods.
Additionally, for the `main_language` method you can provide a list of `additional_token_samplers` to discriminate in case the class `self.token_sampler` produces an ex-aequo.

The reason to modify temporarily the `self.token_sampler` is that it's used by the `proportion_in_each_language` method.

I can see 2 alternatives:
* pass the `token_sampler` to each method as it was before. This is repetitive and kind of defeat the purpose of having the class.
* get rid of the class and leave just the functions. This makes the functions repetitive since you need to pass the same details to every function

adbar: Fair enough, but the original token sampler was not restored in case of early return (`return list_results[0][0]`).

I made another suggestion below, what do you think?

juanjoDiaz:
This looks good to merge for me.

Is there anything left to tackle?

Issue #64, opened 2023-02-08
Ensure functional continuity before releasing v1


adbar:
Due to the refactoring work going on all functions need to be checked once again before releasing a new software version. 

Changes must be documented and ideally old function names could trigger a deprecation warning.

There is one case so far:
- `langdetect` → `language_detector`
see also https://github.com/adbar/simplemma/pull/59#issuecomment-1422395723

juanjoDiaz:
I personally think that it's normal to drop old functions and do breaking changes when introducing mayor version updates.
Even more so when moving from 0.X to 1.0.

In my experience, as the library grows and evolves, it will be impossible to keep forever the backward compatibility.
It is a must to be able to drop things and do major changes or it will become a major blocker for big improvements.

Of course, it is essential to document these changes and how to upgrade.


I'm fine keeping langdetect for this v1.0.
Just letting you know this FYI. 🙂

adbar:
@juanjoDiaz I'll release a new version soon, how could we best document the classes you've added?

juanjoDiaz:
Hi @adbar,

I have 1 change left: to move the different search strategies to a more composable approach and create a class for the lemmatizer to it's consistent with everything else.

I have it almost there but I was pulled to work on other stuff.
Let me try to get a PR ready this week and then we can work on documenting everything.

adbar:
Hi @juanjoDiaz, that's fine, I'll wait for this change before working on a new release.

juanjoDiaz:
This was closed by #90.

Our tests now guarantee continuity.

PR #66, opened 2023-02-09
re-create langdetect() function and add warning

adbar:
see #64


adbar:
This PR will stay open until #65 is resolved.

adbar:
Not necessary anymore.

PR #67, opened 2023-02-09
rules: add generic substitution function

adbar:
There was duplicate code and function type annotations were missing.

Applying the substitution first and potentially return the modified string saves time.


juanjoDiaz (approved):
Good stuff!

PR #69, opened 2023-02-10
Better DE rules

adbar:
Getting rid of unnecessary rules.

The corresponding dictionary needs to be adapted before merging.


simplemma/rules/de.py
```diff
@@ -97,40 +108,25 @@ def apply_de(token: str, greedy: bool = False) -> Optional[str]:
...
-                # lemma identified
-                if not groups:
-                    return token
-                # apply -en/-e/-n/-s patterns
-                return token[: -len(groups[0])]
-            # -end and gerunds
-            if GERUND_DE.search(token):
-                return ENDING_DE.sub("e", token)
     # mostly adjectives and verbs
     elif greedy and token[-1] in ENDING_CHARS_ADJ_DE:
```
juanjoDiaz, line 126: It would be ideal to get rid of this `greedy` here so rules are 100% independent of the greedy search.

An alternative could be to have 2 types of rules:
* `APPLY_RULES`
* `APPLY_GREEDY_RULES`
So the lemmatizer can try the normal rules and then the greedy rules if greedy is set to True.

adbar: I agree, it would be best to simply get rid of the greedy option.

PR #70, opened 2023-02-10
Simplify lemmatizing code

adbar:
The code can be simplified to be more readable.

The prefix search is also moved upward (no incidence on performance for the concerned languages).


simplemma/lemmatizer.py
```diff
@@ -185,31 +198,21 @@ def _return_lemma(
...
-        candidate = APPLY_RULES[lang](token, greedy)
-    # decomposition
-    if candidate is None:  # and greedy is True
-        candidate = _dehyphen(token, datadict, greedy)
-    else:
-        newcandidate = _dehyphen(candidate, datadict, greedy)
-        if newcandidate is not None:
-            candidate = newcandidate
+
+    candidate = (
```
juanjoDiaz, line 202: This looks so much nicer and cleaner!!

Could we wrap the `_affix_search` and the `_greedy_search` in blocks to also fit them here?

simplemma/lemmatizer.py
```diff
@@ -185,31 +198,21 @@ def _return_lemma(
...
+        _dehyphen(token, datadict, greedy)
+        or _simple_search(token, datadict, initial=initial)
+        or apply_rules(token, greedy, lang)
```
juanjoDiaz, line 205: I find weird that `apply_rules` is in the rules modules while `_find_prefixes` is here.

I'd keep both here for consistency.

simplemma/lemmatizer.py
```diff
@@ -151,6 +151,19 @@ def _dehyphen(token: str, datadict: Dict[str, str], greedy: bool) -> Optional[st
     return "".join(splitted)
 
 
+def _find_prefixes(
```
juanjoDiaz, line 154: I'm a big fan of guard clauses to reduce nesting and simplify the code:

```python
def _find_prefixes(
    token: str, lang: Optional[str], datadict: Dict[str, str]
) -> Optional[str]:
    "Subword decomposition: pre-defined prefixes (often absent from vocabulary if they are not words)."
    if lang not in FIND_KNOWN_PREFIXES:
        return None

    prefix = FIND_KNOWN_PREFIXES[lang](token)
    if prefix is None or len(prefix) >= len(token):
        return None

    subword = _simple_search(token[len(prefix) :], datadict)
    if subword is None:
        return None

    return prefix + subword.lower()
```

I know that it repeats the `return None` more times. But reducing the nesting of if clauses really make things simpler to read

simplemma/lemmatizer.py
```diff
@@ -185,31 +198,21 @@ def _return_lemma(
...
     if len(token) <= limit:
         return candidate
-    # subword decomposition: predefined prefixes (absent from vocabulary if they are not words)
-    if candidate is None and lang in FIND_KNOWN_PREFIXES:
-        prefix = FIND_KNOWN_PREFIXES[lang](token)
-        if prefix is not None:
-            subword = _simple_search(token[len(prefix) :], datadict)
-            if subword is not None:
-                candidate = prefix + subword.lower()
+
```
juanjoDiaz, line 215: This is a breaking change. Since words shorter than 6 or 8 characters will know be also tried by the prefix finder.

So, the new version might produce a different result for some tokens.

As you know, I'm happy introducing breaking changes before v1.
But we should think how are we going to handle this in the future.

juanjoDiaz (changes requested):
Looking good!

I have 3 main suggestions:
* Use guard clauses
* Keep `apply_rules` in `lemmatizer.py` for consistency
* Wrap `_affix_search` and `_greedy_search` in function so they can be included in the block of searches

simplemma/lemmatizer.py
adbar: I moved part of the code.

simplemma/lemmatizer.py
adbar: OK, I switched to guard clauses.

simplemma/lemmatizer.py
adbar: Not really but I tried to simplify the rest of the function as well.

adbar:
I'd say the branch is ready to be merged, we can always keep on simplifying later. What do you think @juanjoDiaz?

juanjoDiaz (approved):
I wouldn't have the `apply_rules` and `_find_prefix` methods under `rules/__init__.py` and would just do the lookup from the functions in lemmatizer to skip the extra function call and because the don't add much.

If keeping them, I'd have both of them with the leading underscore or none of them. But not mixed as it is now.

But that's a minor thing.
It's a great improvement overall!

PR #73, opened 2023-02-16
refactor: minimize dictionary lookups in _greedy_search

juanjoDiaz:
Small PR.

This reduces the dictionary lookups from 3 to 1 and remove the counter operations which should improve performance.
Not relevant for a single token but might make a different for a whole text.

Also, correct the logic so zero steps actually means zero.
Before zero steps was doing 1 lookup.


adbar:
Nice, thanks.

PR #74, opened 2023-02-16
refactor: simplify _decompose function

juanjoDiaz:
I haven't modified the comments but I think that they could also use some love

### DIFF
--- a/simplemma/lemmatizer.py
+++ b/simplemma/lemmatizer.py
@@ -79,22 +79,21 @@ def _greedy_search(
 
 def _decompose(
     token: str, datadict: Dict[str, str], affixlen: int = 0
-) -> Tuple[Optional[str], Optional[str]]:
+) -> Optional[str]:
     "Split token into known two known parts and lemmatize the second one."
-    candidate, plan_b = None, None
     # this only makes sense for languages written from left to right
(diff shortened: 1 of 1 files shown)


simplemma/lemmatizer.py
```diff
@@ -106,27 +105,22 @@ def _decompose(
         # try other case
```
adbar, line 105: The section below is not well-covered in terms of tests (my bad), so I'm reluctant to change it. I know the code is experimental and can be improved.

adbar:
The beginning looks good but I'm not sure about all the changes.

simplemma/lemmatizer.py
juanjoDiaz: If you look at lines 110 and 112, it's impossible that candidate is not None. So it makes sense to simply return.
Line 115 is a duplicate of line 100.
And line 121 is followed by a break & return so it makes sense to just return.

So, overall, I didn't change the logic.
Test coverage has increased 0.6% with this change and the only lines that seemed not covered by tests are 108-109 according to the codecov report.

It's up to you but I see this as a low-risk beneficial change 🙂 
And additional tests can always be added.

adbar: I can't think of a good test so it's probably OK to change the lines.
I'll address the issue with the comments another time.

PR #75, opened 2023-02-16
refactor: simplify _dehyphen

juanjoDiaz:
Main thing here is to use `_simple_search` so different capitalizations are considered.

Then, more clearly separating the 3 blocks that compose the search.

### DIFF
--- a/simplemma/lemmatizer.py
+++ b/simplemma/lemmatizer.py
@@ -136,7 +136,7 @@ def _dehyphen(token: str, datadict: Dict[str, str]) -> Optional[str]:
         return None
 
     # try to find a word form without hyphen
-    subcandidate = "".join(t for t in splitted if t not in HYPHENS).lower()
+    subcandidate = "".join([t for t in splitted if t not in HYPHENS]).lower()
     if token[0].isupper():
         subcandidate = subcandidate.capitalize()
     if subcandidate in datadict:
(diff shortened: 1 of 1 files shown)


simplemma/lemmatizer.py
```diff
@@ -134,20 +134,22 @@ def _dehyphen(token: str, datadict: Dict[str, str]) -> Optional[str]:
...
         return None
+
     # try to find a word form without hyphen
-    subcandidate = "".join([t for t in splitted if t not in HYPHENS]).lower()
+    subcandidate = "".join(t for t in splitted if t not in HYPHENS).lower()
     if token[0].isupper():
         subcandidate = subcandidate.capitalize()
-    if subcandidate in datadict:
-        return datadict[subcandidate]
+    subcandidate_lemma = _simple_search(subcandidate, datadict)
```
adbar, line 142: This is more costly and not useful IMO, it looks for different combinations although only "subcandidate" in its current write form is useful.

adbar:
Thanks, I would keep the other changes but revert the one I highlighted above, does that work for you?

adbar:
Ready to merge?

juanjoDiaz:
Looks good to merge!

adbar:
I just restored the list, it's actually faster (not by much though).

Issue #63, opened 2023-02-07
Create configurable Rules Engine


juanjoDiaz:
At the moment we have many rules:

* Simple rules (mainly ending substitutions)
* A quite complex rule system for German
* A rule to decomposed words split by hyphens
* Known prefix recognition for German & Russian
* Affix search used to try identify works skipping the beginning or end of the word (not 100% sure of how this works)
* Greedy search (not sure what this does)

It would be great to have a configurable rule engine.
All this rules can be there as default rules that can be enabled/disabled & users can also add their own rules.

juanjoDiaz:
Thinking more of this...

`en`, `fi`, `nl`, `pl` and `ru` rules are all suffix substitutions with some rules are not being applied if the word doesn't have a minimum length

Where are those length exclusion coming from?
I would help standardizing rules if we just have a minimum word length per language. So it's easy to standardize the rules.

`de`, however, is quite complex. It has different rules for uppercase tokens and for greedy vs non-greedy search. Also the `NOUN_ENDINGS_DE` rules seems a bit complex.

Could we simplify this be more in line with the other languages do?

adbar:
The word length exclusion reduces the impact of the rules in case they are way off: longer words are usually less frequent (so faulty rules will be less apparent) and they are more likely not to be in the dictionary.

I'm working on making the German rules lighter, they are too complex for what they bring.

PR #78, opened 2023-04-09
Refactor/better searches

juanjoDiaz:
Partially related to #63.

The general idea of this PR is to make the lemmatization strategy highly configurable.

To achieve it, the idea is to have:
* a base `LemmatizationStrategy` class
* separate lemmatization strategies for dictionary look ups, dehyphen search, prefix-based search, rule-based search, etc.
* a default strategy that combines the different strategy in the desired way

With this
* users can combine the different strategies in whichever way they want.
* users can create their own extrategies
* users can extend existing strategies

To be honest, I don't love the name `Strategy` but I couldn't think of a better name.

Wdyt?


juanjoDiaz:
The test are working locally.

Any idea of where the issue might be @adbar

osma:
It seems to me that tests are failing on Python 3.6 and 3.7 but passing on 3.8+.

This line:

    from typing import Dict, Optional, Protocol

tries to import the Protocol class, which was introduced in Python 3.8 (see [PEP 544](https://peps.python.org/pep-0544/))
Maybe that could be reason why tests are failing?

juanjoDiaz:
Thanks, @osma !

I forgot that there!
There is a separate issue to introduce Protocol classes but for compatibility reasons we should stick to abstract classes for now 🙂

adbar:
Thanks for the changes! The PR moves a lot of code around at once, I need some time and will look at it next week.

I assume you mostly refactored functions and classes, if there are particular places where you introduced more changes feel free to make a small list so I know where I should pay attention.

juanjoDiaz:
Hi @adbar,

Every strategy under the `strategies` folder is just copy-pasted from the `lemmatizer.py` file.
The changes in the `dictionary_pickler.py` and `language_detector.py` just replace the `_search_lemma` method by a configurable `strategy`.
The change in `lemmatizer.py` creates a configurable class like we did on other parts of the library.
The tests under `rules` folder have been moved to `strategies/defaultrules` and `strategies/defaultprefixes`.
Some tests from `lemmatizer.py` that were just testing a search strategy have been moved under `strategies`.
The tests of `LanguageDetector` and `Lemmatizer` have been adjusted to the new interface.

So, in reality this PR is mainly about moving the strategies to their own folder/files and the new `Lemmatizer` class.

adbar:
There are unused `silent` arguments in `simplemma/lemmatizer.py`, did 
something get lost on the way?

juanjoDiaz:
Hi,

I kept the `silent` argument there for backwards compatibility but I've removed the functionality.
Previous logic was:
```python
    if not silent:
        raise ValueError(f"Token not found: {token}")
    # try to simply lowercase # and len(token) < 10 ?
    return token.lower() if next(iter(dictionaries)) in BETTER_LOWER else token
```

This is something that I wanted to discuss with you.
I think that the correct behaviour is to raise an error and let the customer handle it.

An alternative could be to add this silent option to the `DefaultStrategy`.
Maybe we could add a `falback_strategy` argument to the default strategy that is only called if the `default` strategy doesn't find a lemma for the token. The default `falback_strategy` is to raise the error or to do the lower logic.

Wdyt?

adbar:
Yes that sounds good, let's go for a `fallback_strategy` with the lowercase logic as a default and the error as an option.

adbar:
So we now have a default strategy, I guess we could merge the PR after these steps:
1. Write tests for the new fallback strategies
2. Get rid of the "silent" argument
3. Check eveything for consistency

If it's OK you could tackle (1) & (2) and we would then move on to (3).

adbar:
@juanjoDiaz I'd really like to release a version 1.0.0 soon, if we cannot work on it right now this PR can be merged in v1.0.1 (for example).

juanjoDiaz:
Hi @adbar,

Sorry for the delay. The past few weeks have been hectic on my side.

I just added a test for the fallback strategy, removed the unused "silent" argument and fixed some generic export.

I still would like to release this as v1 to avoid changing the contract once v1 is out.

This week I will have some free most days to look into this so let's aim to release v1 by the end of the week 🙂 .

adbar:
Hi @juanjoDiaz everything looks OK, let's leave this PR open for now in order to take some time to check everything.

Let's aim for a release later this week!
Could you please open a separate PR to describe the changes? In my mind we just need to add a list of the classes you introduced below the rest in the Readme, or maybe just the main ones, something like `ClassName: description`.
You could also add docstrings in the code if you prefer.

juanjoDiaz:
Do you have any library that could generate markdown from Docstring?
I tried lazydocs but it doesn't seem great.

I think that we should document each class an extension point properly if we want people to understand how they work and be able to use them to extend simplemma.

If we just promote the old functions then the general public doesn't gain much from my changes.

adbar:
Sphinx can do it with autodoc I think but it would mean we need to host additional doc pages by readthedocs for example.

juanjoDiaz:
This PR closes #83 and #63

juanjoDiaz:
I just noticed that with this change, the `setup.py` file is broken.
I'm not sure what that file does, so you probably want to look into it.

adbar:
Yes, the setup file is for the optional mypyc compilation but I don't know if anybody uses it.

juanjoDiaz:
Mypy does not support Python 3.6 because it was EoL'd in December 2021.
https://github.com/python/mypy/issues/12640
https://github.com/python/mypy/issues/11974

I don't think that we should support it here either for the same reasons. But at least I will remove mypy from CI with Python 3.6

adbar:
Thanks for the modifications! Let's keep Python 3.6 for v1.0 and drop it soon afterwards.

PR #79, opened 2023-04-09
feat: stricter input by making lang mandatory

juanjoDiaz:
Optional `lang` doesn't make any sense.
We should be stricter


osma:
Makes sense to me :+1:

adbar:
makes sense, thanks!

PR #80, opened 2023-04-09
feat: throw on unsupported languages

juanjoDiaz:
Silently ignore unsupported languages leads to incorrect usage of the library and potential error.

Stricter checks ensure correct usage and should not affect most users.


osma:
Makes sense to me. :+1:

adbar:
That's a good check to have!

PR #81, opened 2023-04-09
feat: get list of supported languages dynamically

juanjoDiaz:
No maintenance needed as languages are added


osma:
Less code to maintain, better named variable, improved test coverage...what's not to like here? Great work! :+1:

adbar:
Just a few minor issues to fix, I'll take care of the merge.

PR #84, opened 2023-05-10
Use Protocol classes

### LINKED ISSUE #71: Add protocol classes for classes that can be extended by the user
At the moment these are:
* DictionaryFactory
* Tokenizer
* TokenSampler

juanjoDiaz:
Closes #71

### DIFF
--- a/simplemma/__init__.py
+++ b/simplemma/__init__.py
@@ -12,6 +12,7 @@ from .lemmatizer import Lemmatizer, lemmatize, lemma_iterator, text_lemmatizer, 
 from .tokenizer import Tokenizer, RegexTokenizer, simple_tokenizer
 from .token_sampler import (
     TokenSampler,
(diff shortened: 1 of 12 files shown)


adbar:
The merging sequence would be #78, then this one, then #85, correct?

Could you please make sure that all the classes use slots? It can make the code run faster at no cost.

adbar:
There are unused imports in dictionary_factory.py and tokenizer.py: ABC.

adbar:
Since these are your changes I think it's best if you resolve the merge conflicts.

adbar:
Thanks for the changes!

What about putting these lines in the `__init__.py` file to avoid duplicate code?
```
import sys

if sys.version_info >= (3, 8):
    from typing import Protocol
else:
    from typing_extensions import Protocol
```

PR #86, opened 2023-05-11
remove optional mypyc compilation

adbar:


osma:
I'm curious what's the reason behind this removal? Was it actually not useful?

osma:
Thanks @adbar, makes a lot of sense!

PR #88, opened 2023-05-23
Ensure __slots__ used everywhere & refactor DictionaryFactory to be part of strategies

juanjoDiaz:
Ensure that all classes use `__slots__` for better performance and put `_` before all the internal properties.


juanjoDiaz:
The second commit refactor DictionaryFactory to be just part of the dictionary-based strategies.

We have now:
- Based on dictionaries:
  - Dictionary lookup (uses dictionary factory)
  - Greedy dictionary lookup (uses dictionary factory)
  - Hyphen removal (uses dictionary lookup strategy)
  - Affix decomposition (uses dictionary lookup strategy & greedy dictionary lookup strategy)
  - Prefix decomposition (only for known prefixes)  (uses dictionary lookup strategy)
- Based on rules
  - Rules

Then the lemmatize & language detector just use these strategies.

Affix decomposition strategy is probably something to review as it it quite hard to understand and it uses greedy search even when `greedy` is set to false, which is counter-intuitive

simplemma/language_detector.py
```diff
@@ -6,32 +6,35 @@
 from .strategies.lemmatization_strategy import LemmatizationStrategy
 from .strategies.default import DefaultStrategy
 
-from .dictionary_factory import DictionaryFactory, DefaultDictionaryFactory
+from .strategies.dictionaries.dictionary_factory import (
```
adbar, line 9: You can also remove this import.

PR #89, opened 2023-05-25
Chore/add docstring

juanjoDiaz:
This includes #88. So, please, merge that first.

PR #90, opened 2023-05-25
Test/add tests for new classes

juanjoDiaz:
It includes #88 and #89. Should be merge after that.

Ensure that we test new classes vs old functions.


adbar:
Thanks!

PR #91, opened 2023-05-26
refactor: provide better strategies exports

juanjoDiaz:
Getting there!

PR #92, opened 2023-05-26
refactor: prefix and rules tests to test the actual strategy and not …

juanjoDiaz:
…the underlying code

Much better testing!
Testing the actual strategy and not the internal details (which are just a trivial lookup anyway...)


adbar:
Please check the formatting (with black), we could also use pre-commit hooks.

juanjoDiaz:
done

juanjoDiaz:
This will work after mergin #91

adbar:
It still doesn't, maybe you need to update something in this branch first.

PR #95, opened 2023-05-26
chore: remove useless logging

juanjoDiaz:
This logging is totally useless for the library.
It's better removed.

I've left logging on the dictionary_pickler since I assume that you use that manually and benefit from the logs


adbar:
OK for the library but why remove the debug level from the tests?

juanjoDiaz:
There was no debug logging form the test.
`logging` was imported and set to debug. However, nothing is logged anywhere.
So, it doesn't do anything.

adbar:
Yes, that's fine.

PR #98, opened 2023-05-29
Feat/better apporach to greedy lookups

juanjoDiaz:
With discussed on another PR that the greedy lookup was used inconsistently across the different strategies.

This is my take on the problem: The default strategy receives a DictionaryLookupStrategy (greedy or not) and uses it for all the other strategies.
So, either everything is greedy or nothing is.

### DIFF
--- a/simplemma/language_detector.py
+++ b/simplemma/language_detector.py
@@ -12,7 +12,12 @@ Functions:
 from operator import itemgetter
 from typing import Dict, List, Tuple, Union
 
-from .strategies import LemmatizationStrategy, DefaultStrategy
+from .strategies import (
+    LemmatizationStrategy,
+    DefaultStrategy,
+    DictionaryLookupStrategy,
+    GreedyDictionaryLookupStrategy,
+)
 from .token_sampler import (
     TokenSampler,
     MostCommonTokenSampler,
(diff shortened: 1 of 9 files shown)


adbar:
Hi @juanjoDiaz this change can have far-reaching consequences for the lemmatization process so I'm going to test it for all languages for which I have reference data to prevent degraded performance.

It appears the recent changes have effectively broken the software... I cannot run my evaluation script so I'd appreciate if you tackle issues #99 and #100 before we move on with this PR.

adbar:
Not sure about the conflicts mentioned by Github, you'd better look into that. I'll run the tests once the `IndexError`  is fixed.

adbar:
@juanjoDiaz I cannot run the evaluation because this branch is out of sync, could you please have a look at it?

In the meantime I'll use the legacy functions which still work.

adbar:
It generally decreases by -0.1 to -0.2, not that much however I'd prefer to keep the functionality as it is. The "greedy" option is actually more like "greedier", the lemmatizer is trying its best even without this option, it's something we could mention in the docs.

I agree your changes make the code simpler but it comes at the cost of deleting a useful component (i.e. try shorter candidates) which gets particularly clear with the tests for German. So I wouldn't go further with this PR. We can let it open to have a look at it later if you wish. Right now stabilizing things before v1 has the priority.

adbar:
I'll now close the PR for the sake of clarity and move on with the rest. We can come back to it later if you find a solution which does not impact accuracy.

PR #101, opened 2023-05-30
fix: bring back initial arg in lemmatize

### LINKED ISSUE #99: Restore `initial` argument in lemmatize
In v0.9.1 it is possible to provide a parameter named `initial` to the `lemmatize()` function. It appears the argument has been deleted in this commit: https://github.com/adbar/simplemma/commit/7c3e833122381c80e1217c8b441a5baa0058252b.

@juanjoDiaz The functionality isn't complete as of now, deletion of the `silent` parameter was fine but I didn't see this one and it turns out I'm actually using it...

juanjoDiaz:
Closes #99.

This brings back intial.
However, I feel strongly that this doesn't belong here as it breaks the single responsibility principle..
`lemmatize` should lemmatize the given token and the decision on the capitalization is a pre-step that depends on how the tokens were obtains.

So, I would rather, keeping it as is.
The same line where you decide if `initial` is true or false, you could use to send `token` or `token.lower()`.

```python
initial = is_initial(my_word)
lemmatize(my_word, my_lang, intiial)
```
is simply

```python
lemmatize(my_word.lower() if is_initial(my_word) else my_word, my_lang)
```

It works the same and it keeps each responsibility where is belongs.

But, at the end of the day, it is your call 🙂

### DIFF
--- a/simplemma/lemmatizer.py
+++ b/simplemma/lemmatizer.py
@@ -63,7 +63,7 @@ def is_known(
 
 
 def lemmatize(
-    token: str, lang: Union[str, Tuple[str, ...]], greedy: bool = False
+    token: str, lang: Union[str, Tuple[str, ...]], greedy: bool = False, initial: bool = False
 ) -> str:
     """Lemmatize a token in the specified language(s).
 
@@ -77,7 +77,7 @@ def lemmatize(
     """
     return Lemmatizer(
         lemmatization_strategy=DefaultStrategy(greedy),
-    ).lemmatize(token, lang)
+    ).lemmatize(token.lower() if initial else token, lang)
 
 
 def text_lemmatizer(
--- a/tests/test_lemmatizer.py
+++ b/tests/test_lemmatizer.py
@@ -557,3 +557,7 @@ def test_get_lemmas_in_text() -> None:
             ".",
         ]
     )
+
+def test_lemmatize() -> None:
+    assert (lemmatize("Dritte", "de", initial=True) == "dritt")
+    assert (lemmatize("Dritte", "de", initial=False) == "Dritter")


adbar:
My bad, I see that you changed it in the [evaluation script](https://github.com/adbar/simplemma/blob/main/training/eval/udscore.py) to `token_form = token["form"].lower() if initial else token["form"]`.

I also see it's not documented and I only used it internally, so it wouldn't be a huge problem.

However it was used by the `lemma_iterator()` function, the new version of it doesn't account for punctuation. Even if it's a debatable point it would mean the new version doesn't reproduce old behavior, here is the old code for reference:

```
def lemma_iterator(
    text: str,
    lang: Optional[Union[str, Tuple[str, ...]]] = None,
    greedy: bool = False,
    silent: bool = True,
    tokenizer: Tokenizer = Tokenizer(),
) -> Iterator[str]:
    """Convenience function to lemmatize a text using a simple tokenizer.
    Returns a list of tokens and lemmata."""
    initial = True
    for token in tokenizer.split_text(text):
        yield lemmatize(token, lang, greedy, silent, initial)
        initial = token in PUNCTUATION
```

juanjoDiaz:
`lemma_iterator()` does take into consideration punctuation exactly as it did before:

It does simply call the Lemmatizer:
https://github.com/adbar/simplemma/blob/1cd7205ef1391d05d5f4b89923ed878daf85af6b/simplemma/lemmatizer.py#L111-L132

Which considers punctuation:
https://github.com/adbar/simplemma/blob/1cd7205ef1391d05d5f4b89923ed878daf85af6b/simplemma/lemmatizer.py#L228-L245

adbar:
Then it's fine, thanks for the prompt answer!

PR #105, opened 2023-06-01
update evaluation and add how-to

adbar:


eval/udscore.py
```diff
@@ -81,44 +78,51 @@
...
-            token_form = token["form"].lower() if initial else token["form"]
+                initial = bool(token["id"] == 1)
+                token_form = token["form"].lower() if initial else token["form"]
 
-            candidate = lemmatizer.lemmatize(token_form, lang=language)
-            greedy_candidate = greedy_lemmatizer.lemmatize(token_form, lang=language)
+                try:
+                    candidate = lemmatizer.lemmatize(token_form, lang=language)
+                except Exception:
+                    candidate = ""
```
juanjoDiaz, line 95: What kind of exception are you getting?
If you get exceptions we should fix simplemma 🙂 

I noticed that tokens that reach the prefix strategy and the whole token match a prefix, throw.
Have you find any other case?

adbar: Yes, it's the IndexError I also saw, do you want to fix it?

adbar: I just made sure the script could run but I'll remove this line as soon as we get it fixed.

PR #106, opened 2023-06-08
fix: issue with tokens that match prefixes

juanjoDiaz:


adbar:
@juanjoDiaz Could you please add a corresponding test case? It can be in any language, just for test coverage and quality assessment.

juanjoDiaz:
done!

PR #108, opened 2023-07-15
feat: use cached object for legacy functions

### LINKED ISSUE #107: Processing time difference between legacy functions and new classes
Even if the existing/legacy functions still work there is a x2.5 time difference between them and the new approach based on your classes @juanjoDiaz. See below for a graphical explanation by pyinstrument (the files are in HTML format, GitHub wouldn't take them without the additional extension).

A lot of time seems to go to classes init each time the lemmatization function is run. If it's the culprit there is probably a way to avoid re-initializing everything each time, what do you think?

[pyinstrument-legacy.html.log](https://github.com/adbar/simplemma/files/11755811/pyinstrument-legacy.html.log)
[pyinstrument-new.html.log](https://github.com/adbar/simplemma/files/11755812/pyinstrument-new.html.log)

juanjoDiaz:
Closes #107 

This is less flexible and potentially consumes more memory. But makes initialization faster if the legacy functions are called many times with the same parameters.

### DIFF
--- a/simplemma/lemmatizer.py
+++ b/simplemma/lemmatizer.py
@@ -19,6 +19,7 @@ from .strategies import (
     LemmatizationStrategy,
     DefaultStrategy,
     DictionaryLookupStrategy,
+    DefaultDictionaryFactory,
     LemmatizationFallbackStrategy,
     ToLowercaseFallbackStrategy,
 )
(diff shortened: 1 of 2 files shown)


adbar:
Yes, that's what I meant, thanks!

PR #109, opened 2023-07-15
docs: add mkdocs page for documentation

juanjoDiaz:
mkdocs page for docs.

I don't know how familiar you are with GitHub Pages, but this can be easily published. See the following:
* https://www.mkdocs.org/user-guide/deploying-your-docs/
* https://squidfunk.github.io/mkdocs-material/publishing-your-site/

I would also be happy to configure that, but then you'd need to trust me with some higher permissions on the repo.

I suggest that we move all content to the page and keep the readme to a minimum with some pointers to the page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,7 +108,6 @@ ENV/
 simplemma/lists/
 
 # Not ready yet
-docs/
 AUTHORS.rst
 CONTRIBUTING.rst
 Makefile
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,0 +1,39 @@
+## Contributions
+
+### Issues, bugs & improvment ideas
+
+Feel free to contribute, notably by [filing issues](https://github.com/adbar/simplemma/issues) for feedback, bug reports, or links to further lemmatization lists, rules and tests.
+
+You can also contribute to this [lemmatization list repository](https://github.com/michmech/lemmatization-lists).
+
+### Contributing to the code
+
+Pull requests are welcome.
+
+Before creating the pull request, please ensure that it pass all the quality checks:
+
+```sh
+# Install dev dependencies
+pip install -r requirements.dev.txt
+# Check code style
+black --check --diff simplemma training tests
+# Check typings
+mypy -p simplemma -p training -p tests
+# Run tests
+pytest --cov=./ --cov-report=xml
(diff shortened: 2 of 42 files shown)


adbar:
Thanks, I agree that moving most of the docs into a separate folder would be better.

Considering hosting I'm going to have a look at the links you provided. So far I've been using `readthedocs.org` for other projects but I don't mind using another approach.

osma:
This looks like a big improvement! But this PR doesn't currently touch the top level README at all, so some content is now duplicated (e.g. the table with supported languages). Shouldn't the sections that have been moved to the docs be removed from the README?

osma:
@adbar 
> We could use links to the relevant sections.

Not sure I understood. Was this directed to me? Should I be more specific in pointing out the duplicated text? Sure, I can do that if it helps. I thought it was quite obvious looking at the PR.

juanjoDiaz:
Hi guys!
Sorry for the delayer response. You got me on vacation!

I agree with @adbar  that the readme should be reduced to a minimum and just point to the Github page hosting the docs.

adbar:
I'd be in favor of sphinx because the [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) function will prove useful to automatically reflect changes made to the functions or classes. I'm open to try mkdocs but I don't know it as well.

As for hosting I've used [readthedocs.org](https://docs.readthedocs.io/en/stable/tutorial/index.html#preparing-your-project-on-github) because it integrates seamlessly with Github but so do Github pages, it's just not so easy I guess.

Since we need to move on I would be in favor of accepting the PR as is and address the other questions later:
1. `mkdocs` or `sphinx` ?
2. `github.io` or `readthedocs.org` ?

@juanjoDiaz @osma What do you think?

juanjoDiaz:
Hi @adbar ,

I think that `github.io` is preferred as it's becoming the standard and it's simpler to keep everything in the same platform.

Regarding `mkdocs` vs `sphinx`, I chose mkdocs because it was easy and I was familiar with it.
I can try to give sphinx a go this week but I can promise that I will be able to complete it.

adbar:
OK then I think I'll just merge the PR as it is and try to get familiar with the process.

Issue #112, opened 2023-08-11
in_target_language can count words twice and return ratios above 1.0


osma:
I noticed that in current `main` version, the `in_target_language` function can return ratios above 1.0 when given more than one language and the words in the input are matching several languages.

Example:

    >>> in_target_language('It was a true gift', lang='en')
    1.0
    >>> in_target_language('It was a true gift', lang='de')
    0.6666666666666666
    >>> in_target_language('It was a true gift', lang=('en','de'))
    1.6666666666666665

Simplemma 0.9.1 doesn't have this problem:

    >>> in_target_language('It was a true gift', lang='en')
    1.0
    >>> in_target_language('It was a true gift', lang='de')
    0.6666666666666666
    >>> in_target_language('It was a true gift', lang=('en','de'))
    1.0

It's not just a question of capping the scores at 1.0. The problem is that a single word can count more than once if it happens to match multiple languages. Below is an example that demonstrates the problem. I added nonsense words that don't match any language, but their presence is compensated by words that are counted twice.

Current `main` version:

    >>> in_target_language('It was a true gift xxx yyy', lang=('en','de'))
    1.0

Simplemma 0.9.1:

    >>> in_target_language('It was a true gift xxx yyy', lang=('en','de'))
    0.6

adbar:
Good catch! That's indeed not the idea here and it's a regression.

adbar:
I believe this line could be changed:
```
results = LanguageDetector(
    lang, token_sampler, DefaultStrategy(greedy)
).proportion_in_each_language(text)
```
→ `proportion_in_target_languages(text)`

However this is at least a performance regression, here is the old code, notice the break which is absent from the new code which makes it less efficient:

```
for l in LANG_DATA:
    candidate = _return_lemma(token, l.dict, greedy=True, lang=l.code)
    if candidate is not None:
        in_target += 1
        break
```

adbar:
@juanjoDiaz Could you please have a look at this issue?

adbar:
The code should now work, do you confirm @osma ?

osma:
I can confirm that it works.

There seems to be an extra line in the test; I commented on that here: https://github.com/adbar/simplemma/pull/114/files#r1327158689

Also, there could be an unit test for this case:

>     >>> in_target_language('It was a true gift xxx yyy', lang=('en','de'))
>     0.6

[...]

Issue #111, opened 2023-08-11
simplemma.lang_detector import no longer working


osma:
I noticed that the [language detection example](https://github.com/adbar/simplemma/blob/main/README.rst#language-detection) in the README is no longer working in current `main` version.

Using simplemma==0.9.1 it works as advertised (although the returned ratios are a bit different from those in the README):

    >>> from simplemma import in_target_language, lang_detector
    >>> lang_detector('"Exoplaneta, též extrasolární planeta, je planeta obíhající kolem jiné hvězdy než kolem Slunce."', lang=("cs", "sk"))
    [('cs', 0.8), ('unk', 0.19999999999999996), ('sk', 0.1)]

But with current `main` version the import of `simplemma.lang_detector` is not working:

    >>> from simplemma import in_target_language, lang_detector
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    ImportError: cannot import name 'lang_detector' from 'simplemma'

It works if I change `lang_detector` to `langdetect`:

    >>> from simplemma import in_target_language, langdetect
    >>> langdetect('"Exoplaneta, též extrasolární planeta, je planeta obíhající kolem jiné hvězdy než kolem Slunce."', lang=("cs", "sk"))
    [('cs', 0.75), ('sk', 0.125), ('unk', 0.25)]

Should the documentation be fixed to correspond with the current naming in the code, or the function `langdetect` renamed back to `lang_detector` so the API remains stable?

adbar:
I guess we can use an alias an import it during init. This was a question we discussed with @juanjoDiaz but something must have got lost around the way.

adbar:
This has already been mentioned in #64.

- The `0.9.1` readme says: `from simplemma.langdetect import in_target_language, lang_detector`
- The current readme says `from simplemma import in_target_language, lang_detector`

So we could go for an alias `lang_detector = langdetect` in the init file or we could simply rename the function.

In any case, the release notes have to mention that the import strategy is slighly different.

adbar:
See also https://github.com/adbar/simplemma/commit/58b3ee7430568738f306a5386085eda6628c47d4:
`from simplemma.langdetect` → `from simplemma.language_detector`

Issue #110, opened 2023-08-07
Plans for simplemma 1.0 release?


osma:
Hi, we are soon going to release [Annif](https://github.com/NatLibFi/Annif) 1.0. We are currently depending on simplemma 0.9.1, which is the last released version, released on January 20th.

Since then a lot has happened in the simplemma `main` branch, including many enhancements by @juanjoDiaz . I've also seen occasional references to a 1.0 release in comments on issues and PRs (for example in #64) and a few issues are tagged with [milestone 1.0](https://github.com/adbar/simplemma/milestone/1).

We would like to know whether it makes sense to wait for another simplemma release (which would enable some new features, thanks to the ability to better control memory usage and caches) or whether to release Annif 1.0 with the current simplemma 0.9.1 dependency. So my questions are:

1. Is the next release going to be simplemma 1.0, or are you still going to do another 0.x release?
2. Do you have any concrete plans for when the next release is going to happen?
3. Is there anything we could do to help with getting out the new release?

Thanks in advance & sorry for asking difficult questions :)

adbar:
Hi @osma, thanks for your feedback, it's an important question indeed. When do you plan to release Annif 1.0?

As far as I know there are at least two (currently open) remaining PRs before releasing version 1. Do you have something else in mind @juanjoDiaz ?

---

Regarding your questions:

1. It is not established the next version has to be 1.0.
2. There are concrete plans but as we are not working full time on this project we are regularly behind schedule anyway.
3. Thanks for asking! You could test the current repository version in your CI/CD environment. and report on bugs if something isn't working as usual. If you wish to use the new classes you could check the documentation and see if you can work with it.
(In your setup you can use `simplemma @ git+https://github.com/adbar/simplemma` as a dependency to get the latest state.)

---

Since everything should be working as it should on the side of already documented funtions I see the following options. This is to make sure you (@osma) can benefit from the latest changes without having to hassle everyone:
- Issue a version 0.9.2 as of now to pass on the changes
- Issue a beta version 1 (not exactly sure how but it's possible) that can be referenced by external users
- Issue a version 1.0.0 even without the full docs and a 1.0.1 with the docs and optional additional improvements

osma:
> When do you plan to release Annif 1.0?

We don't have a set date, but would like to release by the end of August or so. The reason I asked about Simplemma was that we could plan for the release accordingly.

> It is not established the next version has to be 1.0.
[...]
> Issue a version 0.9.2 as of now to pass on the changes

[...]

PR #113, opened 2023-08-11
Add README section on advanced usage via classes

osma:
As discussed in https://github.com/adbar/simplemma/issues/110#issuecomment-1673306133, this PR adds a section to the top level README with examples of advanced usage via the Simplemma classes. I used the cache limiting use case I had in mind for the example, but I tried to explains it as a pattern that can be applied also for other customization requirements. Any comments are welcome, I'm happy to adjust as necessary.

While working on this, I discovered some problems that I reported as separate issues #111 and #112.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -208,6 +208,8 @@ The ``greedy`` argument (``extensive`` in past software versions) triggers use o
 Advanced usage via classes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+*The following classes will be made available in the next version. To start using them, install the latest version from the git repository.*
+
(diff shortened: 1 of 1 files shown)


osma:
One aspect of the class API I'm wondering about is the difference between `Lemmatizer` and `LanguageDetector` in that `Lemmatizer` is language-agnostic (the same instance can be used with many languages/combinations, just passing a different `lang` argument to the `lemmatize()` method) while `LanguageDetector` is given a language (or tuple of languages) when it's constructed, so the same instance cannot be reused if you happen to need a different set of languages.

Neither way is wrong, but it seems like these could perhaps be harmonized - either by making Lemmatizer language-specific, or by making LanguageDetector language-agnostic. Maybe @juanjoDiaz has an explanation for the current situation and whether it makes sense to keep it as it is or to try to unify these.

adbar:
Thanks for the added docs and good point above, you could actually open an issue regarding the harmonization of `Lemmatizer`  and `LanguageDetector`. It's not the priority now though, so we can add a corresponding sentence in the docs if you feel users might fail to understand the current difference.

PR #114, opened 2023-09-12
fix: proportion_in_target_languages not considering tokens present in…

### LINKED ISSUE #112: in_target_language can count words twice and return ratios above 1.0

juanjoDiaz:
… several languages.

Closes #112

### DIFF
--- a/tests/test_language_detector.py
+++ b/tests/test_language_detector.py
@@ -108,6 +108,15 @@ def test_in_target_language() -> None:
         == 1.0
     )
 
+    langs = ("en", "de")
+    text = "It was a true gift"
+    assert (
+        LanguageDetector(lang=langs).proportion_in_target_languages(text)
+        == in_target_language(text, lang=langs)
+        == 1.0
+    )
+    in_target_language("It was a true gift", lang=("en", "de"))
+
 
 def test_main_language():
     text = "Dieser Satz ist auf Deutsch."
--- a/simplemma/language_detector.py
+++ b/simplemma/language_detector.py
@@ -198,14 +198,18 @@ class LanguageDetector:
         Returns:
             float: The proportion of text in the target language(s).
         """
-        return sum(
-            percentage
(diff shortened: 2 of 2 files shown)


adbar:
Thanks!

tests/test_language_detector.py
```diff
@@ -108,6 +108,15 @@ def test_in_target_language() -> None:
...
     )
 
+    langs = ("en", "de")
+    text = "It was a true gift"
+    assert (
+        LanguageDetector(lang=langs).proportion_in_target_languages(text)
+        == in_target_language(text, lang=langs)
+        == 1.0
+    )
+    in_target_language("It was a true gift", lang=("en", "de"))
```
osma, line 118: Why is this here?

PR #119, opened 2024-04-02
update setup

adbar:


adbar:
@juanjoDiaz I'm trying to make a fresh start before working on the rest, do you know why the tests fail?

adbar:
Additional note: on my computer the language detection tests are failing but not the rest.

juanjoDiaz:
Hi @adbar ,

I just rebased #116 and tests are passing just fine.
I also run all tests from main on my local and they worked just fine.
So, the issue has to be something that you have changed in this PR.

All the failing tests exit with code 2, which means that the job was cancelled (probably because some other job failed).
Only [tests (windows-latest, 3.8, true, true)](https://github.com/adbar/simplemma/actions/runs/8522494735/job/23342825833?pr=119#logs) exists with code 1 but doesn't show any different details.

Btw, is there any reason to don't merge outstanding PRs with passing tests before going back into new things?
I recently has to explain to someone how all the classes work and it was quite tedious without the public documentation introduced in #109.

Issue #125, opened 2024-05-14
Function `is_known()` not working as expected (missing word when a rule is active)


adbar:
### Summary

The case happens when a word can be decomposed without being in the dictionary (here because of a rule).

`from simplemma import lemmatize, is_known`

### Legacy

```
>>> lemmatize("Bergungen", lang="de")
'Bergung'
>>> is_known("Bergungen", lang="de")
True
```

### Current state

```
>>> lemmatize("Bergungen", lang="de")
'Bergung'
>>> is_known("Bergungen", lang="de")
False
>>> is_known("Bergungen", lang="de", greedy=True)
False
```

### Trying to debug

```
>>> from simplemma.strategies.dictionary_lookup import DictionaryLookupStrategy
>>> d = DictionaryLookupStrategy()
>>> d.get_lemma("Bergungen", "de")
>>> d.get_lemma("Berge", "de")
'Berg'
```

### Problem

The word "Bergungen" is in the original input list and it can be lemmatized so it is "known". It is already present in the previous package versions. The result should be comparable to "Berge", i.e. something is found.
Also, the greedy option doesn't do anything (although there are rules that can be applied here).

Suggestions:
- Since there is a rule with the words in -ung they are not part of the dictionary file anymore to save some space. So this is an issue with the way language dictionaries are created, or we should apply rules to `is_known()` on top of the dictionary.
- Get rid of `greedy` argument for `is_known()` ?

@juanjoDiaz What do you think?

juanjoDiaz:
Hi @adbar ,

The `is_known` function only look if the word is in dictionary.
This is how it always worked (even before the refactoring to classes).

And you are right. The `greedy` option can be removed as it does nothing.
But that would change as soon as we apply all the the strategies.

We have 2 options:
Option 1: Keep `is_known` and apply all the rules as they are.

Option 2: Remove the `is_known` function from the Lemmatizer class and remove the `fallback_strategy` protocol and it's use in the `lemmatize` method.
This means that, when a lemma is not found, `None` will be returned and the user can decide how to act on it.
At the moment, the default `fallback_strategy` move the word to lowercase for certain languages, although I'm not sure how necessary this is.
We can, of course, keep the legacy `is_known` function.

I like the latter, as it simplifies things.
However, I'm not sure of the impact of removing the lowercasing.

What do you think?

adbar:
[...]

PR #126, opened 2024-05-15
Feat/simplify is known function

juanjoDiaz:
Partially solves #125 

We still need to redo the dictionaries, etc.

### DIFF
--- a/tests/test_lemmatizer.py
+++ b/tests/test_lemmatizer.py
@@ -460,6 +460,9 @@ def test_is_known() -> None:
     assert is_known("FanCY", lang="en") == True
     assert is_known("Fancy-String", lang="en") == False
 
+    assert is_known("espejos", lang=("es", "de")) == True
+    assert is_known("espejos", lang=("de", "es")) == True
+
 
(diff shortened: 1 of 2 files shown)


tests/test_lemmatizer.py
```diff
@@ -452,27 +452,13 @@ def test_subwords() -> None:
...
-        == True
-    )
-    # known words
-    assert (
-        Lemmatizer().is_known("Fancy-String", lang="en")
-        == is_known("Fancy-String", lang="en")
-        == False
-    )
+    assert is_known("FanCY", lang="en") == True
+    assert is_known("Fancy-String", lang="en") == False
```
adbar, line 462: Could you please add a multilingual test just to make sure everything keeps working as intended in the future?

Suggestion: `assert is_known("espejos", lang=("en", "es")) is True`

adbar:
The PR moves `is_known()` out of the Lemmatizer class and removes the greedy argument, all good!

adbar:
Great, thanks.

Issue #72, opened 2023-02-15
Investigate other data structures to store language data


adbar:
So far language data are stored as dictionaries, i.e. key/value pairs of words and ground forms. Python dictionaries are quite efficient in terms of speed and rather efficient space-wise.

However there might be other data structures around which allow for better data compression without impeding speed. Slightly slower code execution would even be acceptable if the amount of necessary RAM decreases.

I thought about using a trie or a graph-like structure (e.g. a DAG) but I haven't found anything that would be directly usable and beneficial. The [Pytries](https://github.com/pytries) group offers interesting options but they would require to use a C library which is a maintenance issue.

At best the data structure could be integrated directly into Simplemma, e.g. a vanilla implementation of a trie in pure Python.

adbar:
Additional note: dictionaries can be more compact if keys and values are `bytes` instead of `str`. This would be a first step to decrease memory footprint.

Dunedan:
I have the use case that I want to detect the language of chat messages in real time and do
lemmatization on them. To do so I obviously have to keep the dictionaries for multiple languages
loaded, but even when I limit it to the most common ones, the memory consumption of the
dictionaries makes it unfeasible.

Therefore, I started looking into more efficient ways to keep the data in memory and ended up
looking into tries as well. To get a feeling for what is possible, I looked for a mature trie
implementation and ended up at [marisa-trie][1], as that library checks most boxes and is one of
the very few Python trie libraries, which is actively maintained.

I then replaced the use of a dicts for the language data with `BytesTrie`, while keeping pickle
and the LZMA-compression and looked at how that affected sizes and loading times. The changes I
made to `simplemma` are available here: https://github.com/Dunedan/simplemma/commit/2bdc2d2e50e38ddf3acfe5a0764faf82bda2817c

|                               | dict    | MARISA-tries with DEFAULT_NUM_TRIES | MARISA-tries with MIN_NUM_TRIES |
|-------------------------------|---------|------------------------------------|--------------------------------|
| Sum file size LZMA compressed | 69.8MB  | 75.0MB                             | 77.6MB                         |
| Sum file sizes uncompressed   | 687.7MB | 191.0MB                            | 221.7MM                        |
| Memory consumption all data   | ~4200MB | ~190MB                             | ~285MB                         |
| Loading time from compressed files for all data     | ~168s   | ~40s                               | ~43s                           |

For the memory consumption I made sure that all dictionaries and not only the 8 most recent ones
are kept in memory.

[...]

PR #128, opened 2024-05-22
use binary strings in dictionaries to save memory

adbar:
See discussions in #72 and #125.

The PR cannot be merged until the dictionaries have been re-processed to account for the change.

### DIFF
--- a/training/dictionary_pickler.py
+++ b/training/dictionary_pickler.py
@@ -5,7 +5,7 @@ import re
 
 from operator import itemgetter
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import ByteString, Dict, List, Optional
 
 import simplemma
 from simplemma.strategies.dictionaries.dictionary_factory import SUPPORTED_LANGUAGES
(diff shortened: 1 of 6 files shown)


training/dictionary_pickler.py
```diff
@@ -140,13 +143,13 @@ def _pickle_dict(
         mydict = dict(sorted(mydict.items(), key=itemgetter(1)))
     if filepath is None:
         filename = f"strategies/dictionaries/data/{langcode}.plzma"
-        filepath = str(Path(simplemma.__file__).parent / filename)
+        filepath = str(Path(__file__).parent.parent / "simplemma" / filename)
```
adbar, line 146: @juanjoDiaz I changed this line in order to store the pickled files locally, they need to be installed afterwards but I thought it's more logical.

adbar: I just added a switch to make the change explicit.

juanjoDiaz:
Sorry for the slow response.

Did you test the performance impact of this?
Is  it worth it for the memory savings?

adbar:
Yes, the difference in performance is either not noticeable or not significant. The package is a little bit lighter even with compressed pickle files so this is obviously good.

PR #131, opened 2024-05-27
prepare version 1.0.0

adbar:
Everyone OK with that @juanjoDiaz @osma ?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -17,11 +17,11 @@ Breaking changes:
 - The `langdetect` submodule is now `language_detector`
 `from simplemma.langdetect import ...` → `from simplemma.language_detector import ...`
 
-Fixes and improvements
+Fixes and improvements:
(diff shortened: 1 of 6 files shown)


Dunedan:
The first paragraph of the "Advanced usage via classes" section in the README.md should probably be deleted as part of a 1.0 release as well:
https://github.com/adbar/simplemma/blob/546ab7ed70e42db6d60bdae6db94efb493cb1421/README.md?plain=1#L222-L224

osma:
I haven't tested the codebase in a while (still using 0.9.1 in Annif) but I will try to do so. Not before next week though.

adbar:
I guess it was about time... Things should be pretty stable, if there are problems we can just fix them and release a new version.

juanjoDiaz:
I'm a bit late but glad to see this out.
As you mentioned, we can always do more release.

One thing that I'm missing is to add the documentation: https://adbar.github.io/simplemma
To the repo front page so it's accesible to everyone 🙂

PR #133, opened 2024-05-30
Add a dictionary factory backed by MARISA-tries

Dunedan:
This adds an additional dictionary factory backed by MARISA-tries. This dictionary factory on average offers 20x lower memory usage and 100x faster initialization time, in exchange for reduced lemmatization and language detection performance.

The first time loading a dictionary with the `TrieDictionaryFactory` requires more memory and will take a few seconds, as the trie-backed dictionary has to be generated on-the-fly from the pickled dict-based dictionary first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,61 @@ LANG_CACHE_SIZE = 5  # How many language dictionaries to keep in memory at once 
 For more information see the
 [extended documentation](https://adbar.github.io/simplemma/).
 
+### Reducing memory usage
+
+For situations where low memory usage and fast initialization time are
+more important than lemmatization and language detection performance,
+Simplemma ships another `DictionaryFactory`, which uses a trie as
+underlying data structure instead of a Python dict.
+
+Using the `TrieDictionaryFactory` reduces memory usage on average by
+20x and initialization time by 100x, but comes at the cost that
+performance can be down 50% or even more compared to what Simplemma
+otherwise achieves, depending on the specific usage.
+
+To use the `TrieDictionaryFactory` you have to install Simplemma with
+the `marisa-trie` extra dependency:
+
+```
+pip install simplemma[marisa-trie]
+```
+
+Then you have to create a custom strategy using the
+`TrieDictionaryFactory` and use that for `Lemmatizer` and
+`LanguageDetector` instances:
+
+``` python
+>>> from simplemma import LanguageDetector, Lemmatizer
+>>> from simplemma.strategies import DefaultStrategy
+>>> from simplemma.strategies.dictionaries import TrieDictionaryFactory
+
+>>> lemmatization_strategy = DefaultStrategy(dictionary_factory=TrieDictionaryFactory())
+
+>>> lemmatizer = Lemmatizer(lemmatization_strategy=lemmatization_strategy)
+>>> lemmatizer.lemmatize('doughnuts', lang='en')
+'doughnut'
+
+>>> language_detector = LanguageDetector('la', lemmatization_strategy=lemmatization_strategy)
+>>> language_detector.proportion_in_target_languages("opera post physica posita (τὰ μετὰ τὰ φυσικά)")
+0.5
+```
+
+While memory usage and initialization time when using the
+`TrieDictionaryFactory` are significantly lower compared to the
+`DefaultDictionaryFactory`, that's only true if the trie dictionaries
+are available on disk. That's not the case when using the
+`TrieDictionaryFactory` for the first time, as Simplemma only ships
+the dictionaries as Python dicts. The trie dictionaries have to be
+generated once from the Python dicts. That happens on-the-fly when
+using the `TrieDictionaryFactory` for the first time for a language and
+will take a few seconds and use as much memory as loading the Python
+dicts for the language requires. For further invocations the trie
+dictionaries get cached on disk.
+
+If the computer supposed to run Simplemma doesn't have enough memory to
+generate the trie dictionaries, they can also be generated on another
+computer with the same CPU architecture and copied over to the cache
+directory.
 
 ## Supported languages
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,8 @@
 black==24.4.2
 flake8==7.0.0
+marisa_trie==1.2.0
 mypy==1.10.0
+platformdirs==4.2.2
 pytest==8.2.1
 pytest-cov==5.0.0
 types-requests==2.32.0.20240523
--- a/simplemma/strategies/dictionaries/dictionary_factory.py
+++ b/simplemma/strategies/dictionaries/dictionary_factory.py
@@ -14,7 +14,7 @@ from abc import abstractmethod
 from functools import lru_cache
 from os import listdir, path
 from pathlib import Path
-from typing import ByteString, Dict, Protocol
+from typing import ByteString, Dict, Mapping, Protocol
 
 DATA_FOLDER = str(Path(__file__).parent / "data")
 SUPPORTED_LANGUAGES = [
@@ -62,7 +62,7 @@ class DictionaryFactory(Protocol):
     def get_dictionary(
         self,
         lang: str,
-    ) -> Dict[ByteString, ByteString]:
+    ) -> Mapping[str, str]:
         """
         Get the dictionary for a specific language.
 
@@ -70,12 +70,31 @@ class DictionaryFactory(Protocol):
             lang (str): The language code.
 
         Returns:
-            Dict[str, str]: The dictionary for the specified language.
+            Mapping[str, str]: The dictionary for the specified language.
 
         Raises:
             ValueError: If the specified language is not supported.
         """
         raise NotImplementedError
+
+
+class MappingStrToByteString(Mapping[str, str]):
+    """Wrapper around ByString dict to make them behave like str dict."""
+
+    __slots__ = ["_dict"]
+
+    def __init__(self, dictionary: Dict[bytes, bytes]):
+        self._dict = dictionary
+
+    def __getitem__(self, item: str):
+        return self._dict[item.encode()].decode()
+
+    def __iter__(self):
+        for key in self._dict:
+            yield key.decode()
+
+    def __len__(self):
+        return len(self._dict)
 
 
 class DefaultDictionaryFactory(DictionaryFactory):
@@ -86,7 +105,7 @@ class DefaultDictionaryFactory(DictionaryFactory):
     It provides functionality for loading and caching dictionaries from disk that are included in Simplemma.
     """
 
-    __slots__ = ["_data", "_load_dictionary_from_disk"]
+    __slots__ = ["_load_dictionary_from_disk"]
 
     def __init__(self, cache_max_size: int = 8):
         """
@@ -96,7 +115,6 @@ class DefaultDictionaryFactory(DictionaryFactory):
             cache_max_size (int): The maximum size of the cache for loaded dictionaries.
                 Defaults to `8`.
         """
-        self._data: Dict[str, Dict[ByteString, ByteString]] = {}
         self._load_dictionary_from_disk = lru_cache(maxsize=cache_max_size)(
             _load_dictionary_from_disk
         )
@@ -104,7 +122,7 @@ class DefaultDictionaryFactory(DictionaryFactory):
     def get_dictionary(
         self,
         lang: str,
-    ) -> Dict[ByteString, ByteString]:
+    ) -> Mapping[str, str]:
         """
         Get the dictionary for a specific language.
 
@@ -112,11 +130,11 @@ class DefaultDictionaryFactory(DictionaryFactory):
             lang (str): The language code.
 
         Returns:
-            Dict[str, str]: The dictionary for the specified language.
+            Mapping[str, str]: The dictionary for the specified language.
 
         Raises:
             ValueError: If the specified language is not supported.
         """
         if lang not in SUPPORTED_LANGUAGES:
             raise ValueError(f"Unsupported language: {lang}")
-        return self._load_dictionary_from_disk(lang)
+        return MappingStrToByteString(self._load_dictionary_from_disk(lang))
--- a/simplemma/strategies/dictionaries/trie_directory_factory.py
+++ b/simplemma/strategies/dictionaries/trie_directory_factory.py
@@ -1,0 +1,123 @@
+import logging
+from collections.abc import MutableMapping
+from functools import lru_cache
+from pathlib import Path
+from typing import ByteString, Dict, List, Optional, cast
+
+from marisa_trie import BytesTrie, HUGE_CACHE  # type: ignore[import-not-found]
+from platformdirs import user_cache_dir
+
+from simplemma import __version__ as SIMPLEMMA_VERSION
+from simplemma.strategies.dictionaries.dictionary_factory import (
+    DefaultDictionaryFactory,
+    DictionaryFactory,
+    SUPPORTED_LANGUAGES,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class TrieWrapDict(MutableMapping):
+    """Wrapper around BytesTrie to make them behave like dicts."""
+
+    def __init__(self, trie: BytesTrie):
+        self._trie = trie
+
+    def __getitem__(self, item):
+        return self._trie[item.decode()][0]
+
+    def __setitem__(self, key, value):
+        raise NotImplementedError
+
+    def __delitem__(self, key):
+        raise NotImplementedError
+
+    def __iter__(self):
+        for key in self._trie.iterkeys():
+            yield key.encode()
+
+    def __len__(self):
+        return len(self._trie)
+
+
+class TrieDictionaryFactory(DictionaryFactory):
+    """Memory optimized DictionaryFactory backed by MARISA-tries.
+
+    This dictionary factory creates dictionaries, which are backed by a
+    MARISA-trie instead of a dict, to make them consume very little
+    memory compared to the DefaultDictionaryFactory. Trade-offs are that
+    lookup performance isn't as good as with dicts.
+    """
+
+    __slots__: List[str] = []
+
+    def __init__(
+        self,
+        cache_max_size: int = 8,
+        use_disk_cache: bool = True,
+        disk_cache_dir: Optional[str] = None,
+    ) -> None:
+        """Initialize the TrieDictionaryFactory.
+
+        Args:
+            cache_max_size (int): The maximum number dictionaries to
+                keep in memory. Defaults to `8`.
+            use_disk_cache (bool): Whether to cache the tries on disk to
+                speed up loading time. Defaults to `True`.
+            disk_cache_dir (Optional[str]): Path where the generated
+                tries should be stored in. Defaults to a Simplemma-
+                specific subdirectory of the user's cache directory.
+        """
+
+        if disk_cache_dir:
+            self._cache_dir = Path(disk_cache_dir)
+        else:
+            self._cache_dir = (
+                Path(user_cache_dir("simplemma")) / "marisa_trie" / SIMPLEMMA_VERSION
+            )
+        self._use_disk_cache = use_disk_cache
+        self._get_dictionary = lru_cache(maxsize=cache_max_size)(
+            self._get_dictionary_uncached
+        )
+
+    def _create_trie_from_pickled_dict(self, lang: str) -> BytesTrie:
+        """Create a trie from a pickled dictionary."""
+        unpickled_dict = DefaultDictionaryFactory(cache_max_size=0).get_dictionary(lang)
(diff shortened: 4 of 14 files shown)


README.md
```diff
@@ -260,6 +260,61 @@ LANG_CACHE_SIZE = 5  # How many language dictionaries to keep in memory at once
...
+`TrieDictionaryFactory` are significantly lower compared to the
+`DefaultDictionaryFactory`, that's only true if the trie dictionaries
+are available on disk. That's not the case when using the
+`TrieDictionaryFactory` for the first time, as Simplemma only ships
+the dictionaries as Python dicts. The trie dictionaries have to be
+generated once from the Python dicts. That happens on-the-fly when
+using the `TrieDictionaryFactory` for the first time for a language and
+will take a few seconds and use as much memory as loading the Python
+dicts for the language requires. For further invocations the trie
+dictionaries get cached on disk.
```
osma, line 312: Would it be helpful to explain where the cache is located on disk?

osma:
Awesome work @Dunedan !

adbar:
Thanks for sharing your insights and for contributing this functionality!

Everything looks good at first sight but we need to update the installation process on Github actions.

adbar:
@Dunedan Your PR only works on Windows, there seems to be an issue with the `.dic` file names, could you please look into it?

There are also lines not covered by the tests, it should be easy to add further tests.

Dunedan:
> @Dunedan Your PR only works on Windows, there seems to be an issue with the `.dic` file names, could you please look into it?

I guess you're talking about the tests, which only succeed on the Windows runners so far. At first glance that looks like a list ordering issue affecting just the test assertions. Should be easy to fix. I'll probably do so tomorrow.

If you're not talking about the tests, please let me know and provide a bit more details about what functionality doesn't work and how that manifests. I'd be surprised about that though, as I did develop and test the code with Linux and didn't encounter any problems there.

> There are also lines not covered by the tests, it should be easy to add further tests.

Sure, no big deal, I can add tests for that as well.

requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 black==24.4.2
 flake8==7.0.0
+marisa_trie==1.1.1
```
juanjoDiaz, line 3: Shouldn't `platformdirs` also be included?

simplemma/strategies/dictionaries/trie_directory_factory.py
```diff
@@ -0,0 +1,161 @@
+import hashlib
+import logging
+from collections.abc import MutableMapping
+from functools import lru_cache
+from pathlib import Path
+from typing import ByteString, Dict, List, Optional, cast
+
+from marisa_trie import BytesTrie, HUGE_CACHE  # type: ignore[import-not-found]
+from platformdirs import user_cache_dir
```
juanjoDiaz, line 9: `platformdirs` is also a external dependency. So, why it doesn't need the type ignore like `marisa_trie`?

simplemma/strategies/dictionaries/trie_directory_factory.py
```diff
@@ -0,0 +1,161 @@
...
+    MARISA-trie instead of a dict, to make them consume very little
+    memory compared to the DefaultDictionaryFactory. Trade-offs are that
+    lookup performance isn't as good as with dicts.
+    """
+
+    __slots__: List[str] = []
+
+    def __init__(
+        self,
+        cache_max_size: int = 8,
```
juanjoDiaz, line 56: We are doing double caching: in-memory and in-disk.
These params make unclear what it is about.

simplemma/strategies/dictionaries/trie_directory_factory.py
```diff
@@ -0,0 +1,161 @@
...
+        hasher = hashlib.sha256()
+        with self._get_pickled_dict_path(lang).open("rb") as fp:
+            while True:
+                chunk = fp.read(4096)
+                if not chunk:
+                    break
+                hasher.update(chunk)
+        return hasher.hexdigest()
+
+    def _cache_is_current(self, lang: str) -> bool:
```
juanjoDiaz, line 96: What's the scenario in which the in-disk trie won't match the in-memory one?

simplemma/strategies/dictionaries/trie_directory_factory.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class TrieWrapDict(MutableMapping):
+    """Wrapper around BytesTrie to make them behave like dicts."""
+
+    def __init__(self, trie: BytesTrie):
+        self._trie = trie
+
+    def __getitem__(self, item):
+        return self._trie[item.decode()][0]
```
juanjoDiaz, line 27: Having to do this decoding after just encoding feels a bit wasteful, every time makes me think if we did a right change changing this to bytestring...

tests/strategies/dictionaries/test_trie_dictionary_factory.py
```diff
@@ -0,0 +1,234 @@
+from collections.abc import ItemsView, KeysView
```
juanjoDiaz, line 1: Could we somehow reuse generic `DictionaryFactory` tests for all the implementations?

juanjoDiaz (changes requested):
I've dropped a bunch of comments.
I like the idea of using tries but I'm not fully convinced of the implementation.
I think that we should have a `TrieDictionaryFactory` which:
* has a method to convert all (or the selected) pickled dictionaries into pickled tries and save them on disk
* has a method to load a dictionary which loads the dicto from disk if there or from the original dictionary as needed
* doesn't do all that hashing logic because dictionaries are not likely to change unless the library is updated
* reconsider if the current bytestring usage is right

README.md
Dunedan: I tried not to document every little detail in the README to avoid blowing it up too much. IMO there should be a separate API-documentation for Simplemma to cover stuff like that. However, if it's desired, please let me know and I'll happily add it.

simplemma/strategies/dictionaries/trie_directory_factory.py
```diff
@@ -0,0 +1,161 @@
...
+from simplemma.strategies.dictionaries.dictionary_factory import (
+    SUPPORTED_LANGUAGES,
+    DictionaryFactory,
+    _load_dictionary_from_disk,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class TrieWrapDict(MutableMapping):
```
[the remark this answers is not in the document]
Dunedan: One of the constraints I gave myself was to implement this functionality without requiring changes to the existing code of Simplemma. That's why I didn't modify expected return types of the `DictionaryFactory` for example. Doing so would certainly simplify things, however that's something I didn't want to decide on my own.

It's a `MutableMapping` right now, because a `dict`, what's used when using the `DefaultDictionaryFactory`, is as well. No further reason beyond that.

simplemma/strategies/dictionaries/trie_directory_factory.py
```diff
@@ -0,0 +1,161 @@
...
+    memory compared to the DefaultDictionaryFactory. Trade-offs are that
+    lookup performance isn't as good as with dicts.
+    """
+
+    __slots__: List[str] = []
+
+    def __init__(
+        self,
+        cache_max_size: int = 8,
+        use_disk_cache: bool = True,
```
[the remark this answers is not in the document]
Dunedan: Right now if `disk_cache_dir` is `None`, a subdirectory in the users platform-specific cache directory is used to store the cache. So `use_disk_cache` is used right now to distinguish between disabling the disk cache and using the default cache location.

requirements-dev.txt
Dunedan: I guess @adbar just forgot it when adjusting the `requirements-dev.txt`. I just added it there as well.

simplemma/strategies/dictionaries/trie_directory_factory.py
Dunedan: While I'm not entirely sure, my guess is that mypy can't get any type information, because marisa_trie isn't Python code, but a C-extension and doesn't provide type information.

simplemma/strategies/dictionaries/trie_directory_factory.py
Dunedan: I completely agree. I just kept that parameter from `DefaultDictionaryFactory` to be able to use `TrieDictionaryFactory` as a drop-in replacement for it.

simplemma/strategies/dictionaries/trie_directory_factory.py
Dunedan: This method is for checking whether the tries cached on disk still match the pickled bytestring dictionaries. They don't match anymore whenever a new version of Simplemma gets released which includes updated pickled bytestring dictionaries, resulting in the need to regenerate the cached tries.

simplemma/strategies/dictionaries/trie_directory_factory.py
Dunedan: While I somewhat agree, it wasn't much better before, as `BytesTrie` expects the keys to be strings and the values to be bytestrings. So while now the keys have to be encoded/decoded, before the values had to be.

tests/strategies/dictionaries/test_trie_dictionary_factory.py
Dunedan: For common functionality that would be helpful and I considered implementing dictionary agnostic tests, but concluded that this would be out-of-scope for this PR.

requirements-dev.txt
adbar: Yes, that was the issue, thanks.

adbar:
Python 3.6 and 3.7 are now deprecated and the token error with 3.10 on MacOS is a Codecov problem, nothing serious.

Thanks for your work @Dunedan !

@juanjoDiaz Can we merge the PR or do you have other questions?

juanjoDiaz:
I'm all in for having trie-backed dictionaries.

However, I'm still not convinced of the implementation.
I did a quick and dirty alternative proposal (not fully tested, just to show my thinking)

Dunedan:
> I'm all in for having trie-backed dictionaries.
> 
> However, I'm still not convinced of the implementation. I did a quick and dirty alternative proposal (not fully tested, just to show my thinking)

Thanks for that alternative proposal. :+1:

As I'd prefer to keep the discussion in a single place, I'm going to comment on your PR here:

> - It encapsulates better the bytestring optimization while keeping the dictionary factory working on strings

That's what I'd prefer as well, but as that requires changes to the `DictionaryFactory` protocol, I didn't include it in my PR so far. If changing the `DictionaryFactory` protocol is up for discussion, I'm all for it. :+1:

> - It does not require a external dependency to pick a temp folder for disk caching

The way you implemented that has several disadvantages from my point of view:

- in cases where no per-user/per-process temp directory is used (e.g. when not using systemd's `PrivateTmp=true` option), the default cache directory would be shared by all local users. Depending on user groups and umasks, only the first user would be able to write in there. This could be solved though by adding the name or uid of the current user as part of the file path, like other applications do as well.
- the temp directory is often backed by memory (e.g. through tmpfs), therefore it's no good place to store data, which should be cached on disk only
- the generated trie dictionaries would be regenerated on every reboot and depending on the configuration for temp directory and usage of simplemma even more frequently (e.g. by default systemd deletes temporary files which haven't been accessed for 10 days)

All of that is why I didn't use a temp directory, but a user-specific cache directory instead.

> - It does not require all the additional hashing logic. It simply caches data in a folder specific to simplema's version or wherever the user says

In the end that's balancing between minimizing the number of times the trie-based dictionaries have to be regenerated and implementation complexity. I (obviously) opted for minimizing the number of regenerations, as I expect Simplemma not to update dictionaries with every release and regenerating them is pretty costly in terms of CPU and memory usage. If Simplemma is going to update dictionaries more frequently than I anticipate, then the hashing wouldn't make much sense of course.

> - It does not use internal functions of simplemma like _load_dictionary_from_disk

I like that. :+1:

adbar:
I guess it would now be best to integrate some of the changes discussed into this PR? We could first start with the changes you both agree on.

Besides, as far as I'm concerned we can make changes in the `DictionaryFactory` protocol.

juanjoDiaz:
We agree on modifying the `DictionaryFactory` protocol.
I cando that in my other PR.
And then the trie implementation could be adjusted to that.

I see 3 aspects that we need to align:
* How to cache on disk
* Is hashing needed to notice if dictionaries changed?
* should this be part of `simplemma` or published as a separate module?

### How to cache on disk

> The way you implemented that has several disadvantages from my point of view:
>
> * in cases where no per-user/per-process temp directory is used (e.g. when not using systemd's PrivateTmp=true option), the default cache directory would be shared by all local users. Depending on user groups and umasks, only the first user would be able to write in there. This could be solved though by adding the name or uid of the current user as part of the file path, like other applications do as well.
>
> * the temp directory is often backed by memory (e.g. through tmpfs), therefore it's no good place to store data, which should be cached on disk only
>
> * the generated trie dictionaries would be regenerated on every reboot and depending on the configuration for temp directory and usage of simplemma even more frequently (e.g. by default systemd deletes temporary files which haven't been accessed for 10 days)

The more I think about this, the more I think that we should simple store the trie dictionaries in `str(Path(__file__).parent / data_marisa_tries`. It will be there for all runs of simplemma no matter how many processes you have running simplemma.
We could, of course, have an optional argument to provide a custom path. The same could also be used in the `DefaultDictionary` to provide own pickled dictionaries.

I think that we need to think of how Simplemma is used.
I expect that simplemma is use either in scripts or as part of a server.
In both cases, I would simply generate the tries on disk before running the script or starting the server.
If I put my server in a docker container, I would build the tries during startup or even doing the image creation.
I wouldn't expect much dynamism on creating the tries.

That's also why I propose to have a method that allow to pre-generate all tries in disk which can be a simple for loop documented.

### Is hashing needed to notice if dictionaries changed?

> In the end that's balancing between minimizing the number of times the trie-based dictionaries have to be regenerated and implementation complexity. I (obviously) opted for minimizing the number of regenerations, as I expect Simplemma not to update dictionaries with every release and regenerating them is pretty costly in terms of CPU and memory usage. If Simplemma is going to update dictionaries more frequently than I anticipate, then the hashing wouldn't make much sense of course.

I think that Simplemma never updates the dictionaries because it doesn't have a fully automated mechanism to do so.
Otherwise, I would update the dictionaries with every single release of Kaikki or whatever other dataset that we want to use. That's the reason why I would like to really automate the dictionary training.

So, for me the question is: are we going to release sooo many versions of simplemma that regenerating the tries with every trie would become a problem? I don' t think so.

And I have the same remark as before. How will simplemma be used?
In servers and APIs, I would consider creating the trie as part of setting up the service (building the docker image, building the VM or manually setting up the server, depending on how modern you are).

### Should this be part of `simplemma` or published as a separate module?

One of the beauties of simplemma is that it's dependency free.
Having this would make it depend, even if only for one optional functionality, on an external library.

Also, today is marisa trie. But tomorrow it could be another trie implementation, a better hashmap or something else.
Do we want to add all those to simplemma?
Or should they be published as external modules that just implement the protocol so can be used by simplemma easily?

adbar:
I would also prefer to integrate it into simplemma, as long as the installation remains optional the package is still usable without dependencies.

As for the rest of the questions I'll keep on following the discussion as I have to fixed opinion on this. Maybe keep hashing on the side for now for the sake of simplicity.

juanjoDiaz:
Alright, we agreed on:
* Introducing my changes to the `DictionaryFactory` to simplify things
* Keeping the trie implementation within Simplemma
* Not introducing hashing for now

The only thing left is to agree on the disk caching.
I'm no t really against it. I'm just trying to keep things simple.
I'm happy to keep the high configurability of the location with a sensible default.
However, I'd like to remove the dependency on `platformdirs` if possible.

Dunedan:
> Alright, we agreed on:
> 
> * Introducing my changes to the `DictionaryFactory` to simplify things
> * Keeping the trie implementation within Simplemma
> * Not introducing hashing for now

All of that is already implemented in this PR. :slightly_smiling_face: 

> The only thing left is to agree on the disk caching. I'm no t really against it. I'm just trying to keep things simple. I'm happy to keep the high configurability of the location with a sensible default. However, I'd like to remove the dependency on `platformdirs` if possible.

IMO the most simple and sensible solution is to use the users cache directory as default and get its location using `platformdirs`. If you disagree, please suggest a better alternative.

juanjoDiaz:
I would use https://docs.python.org/3/library/tempfile.html instead of `platformdirs` to have a default temp folder with no dependencies.
And let the user provide it's own folder if he wants more control.

Dunedan:
> I would use https://docs.python.org/3/library/tempfile.html instead of `platformdirs` to have a default temp folder with no dependencies.

So this is the same approach you implemented in your PR and I above already explained the downsides this approach has. To sum it up: You prefer to have a default configuration which keeps all generated tries in memory and regenerates them on every reboot (or potentially more often), just to avoid having to use an additional, tiny, well maintained, pure-Python dependency. That approach negates the lower memory benefit of using tries and slows down the users applications as the tries have to be regenerated regularly. For long-running applications using Simplemma which get started on boot, not caching tries would even be better than this approach, as it'd mean lower memory utilization and the tries wouldn't have to be regenerated more often.

In my opinion that's not a user friendly default and I still don't understand the insistence on not using `platformdirs`. If you prefer adding additional code to Simplemma instead of adding an additional dependency, including a copy of the relevant code from `platformdirs` would be an alternative option, still providing the benefit of a sensible caching location.

adbar:
Since this PR entails substantial work I'd be in favor of moving on with the integration. We can always amend it later to improve on the issues discussed here.

@Dunedan The tests do not pass for Python 3.8, could you please have a look at it?

Dunedan:
> @Dunedan The tests do not pass for Python 3.8, could you please have a look at it?

Sorry, I missed that. This should be fixed now.

If you're fine with the code as-is, I suggest I rebase & squash the commits in this branch to have a single commit for all the changes and promote the pull request from "Draft" to "Ready for review" afterwards.

juanjoDiaz:
> So this is the same approach you implemented in your PR and I above already explained the downsides this approach has. To sum it up: You prefer to have a default configuration which keeps all generated tries in memory and regenerates them on every reboot (or potentially more often), just to avoid having to use an additional, tiny, well maintained, pure-Python dependency. That approach negates the lower memory benefit of using tries and slows down the users applications as the tries have to be regenerated regularly. For long-running applications using Simplemma which get started on boot, not caching tries would even be better than this approach, as it'd mean lower memory utilization and the tries wouldn't have to be regenerated more often.

Using `tempfile` does not keep things in memory but on a temp folder in disk.
My insistence in not having `platformdirs` is that it's an external dependency that the user will know that must be installed for things to work.
The only difference that I can see with `platformdirs` is that `tempfile` gives a folder per process whereas `platformdirs` gives a generic folder shared by every process.
I think that it makes sense to have temp data scoped by process.
Lazy caching by many processes, if not done carefully may produce race conditions.

Dunedan:
> Using `tempfile` does not keep things in memory but on a temp folder in disk.

On Linux `tempfile` creates files and directories in `/tmp` which is nowadays often backed by `tmpfs`, which stores its contents in memory.

> My insistence in not having `platformdirs` is that it's an external dependency that the user will know that must be installed for things to work.

I don't know how you install Python packages, but usually the user doesn't need to know about the dependencies of a package. In this case the user has to just follow the instructions for using tries and install `simplemma[marisa-trie]` e.g. with pip:
```
pip install simplemma[marisa-trie]
```

`platformdirs` will be installed automatically as part of that.

> The only difference that I can see with `platformdirs` is that `tempfile` gives a folder per process whereas `platformdirs` gives a generic folder shared by every process. I think that it makes sense to have temp data scoped by process. Lazy caching by many processes, if not done carefully may produce race conditions.

As explained already, there are more than just this single difference.
It'd be easy to use a per-process sub-directory in a users cache directory, however that'd remove the benefit of caching, as they'd have to be regenerated on every invocation of the application.

Regarding race conditions: There is one possible race condition right now: If two processes start in parallel, one doesn't find a cached trie, starts generating it and is still in the process of writing it to disk when the second process tries to reads the unfinished file from the cache, then the second process will retrieve an exception, because the not-yet-finished-file isn't valid. I expect that case to be very rare though and fixing it would be as easy as writing the cached trie to a temporary file and moving it to its final position after writing.

simplemma/strategies/dictionaries/dictionary_factory.py
```diff
@@ -15,6 +15,7 @@
 from os import listdir, path
 from pathlib import Path
 from typing import ByteString, Dict, Protocol
+from typing import Mapping
```
adbar, line 18: This line could be merged with the one above.

adbar:
@Dunedan Thanks for fixing the PR, I believe you can prepare the PR so that it can be merged.

(My understanding is that merging the PR is not the end of the discussion and that work on this optional function can continue beyond this point.)

adbar:
Thanks!

PR #134, opened 2024-06-05
Drop support for Python 3.6 and 3.7

Dunedan:
Python 3.7 is end-of-life for a year now and Python 3.6 even longer, so let's drop support for them to not limit the use of Python language features to one supported by those outdated Python versions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,12 +19,6 @@ jobs:
         os: [ubuntu-latest]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13-dev']
         include:
-        # custom tests
-        # other OS version necessary
-        - os: ubuntu-20.04
-          python-version: '3.6'
-        - os: ubuntu-20.04
-          python-version: '3.7'
         # common versions on MacOS
         - os: macos-latest
           python-version: '3.10'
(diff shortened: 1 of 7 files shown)


adbar:
Yes I was thinking about it too, I know people with old computers and there are downloads for older Python version but maybe it's just tests. Version 1 is out now and can be used in such cases.

Two suggestions:
- You can search the code for "3.6"/"3.7" mentions and remove the lines accordingly.
- Since we're at it we could modernize the setup, I'm thinking about `pyproject.toml` + `setuptools`, this would be another PR.

Dunedan:
> Yes I was thinking about it too, I know people with old computers and there are downloads for older Python version but maybe it's just tests. Version 1 is out now and can be used in such cases.

People using old Python version will inevitably have to pick old versions of libraries, as most Python libraries only support still supported Python versions. IMO it's not worth the additional effort to continue to support unsupported Python versions.

> * You can search the code for "3.6"/"3.7" mentions and remove the lines accordingly.

I had done that already and to my knowledge there are no mentions of these Python versions anywhere in the project anymore. What I missed were the conditional imports (`if sys.version_info >= (3, 8) …`), but I removed those as well now.

> * Since we're at it we could modernize the setup, I'm thinking about `pyproject.toml` + `setuptools`, this would be another PR.

Yes, I'd argue as well that this should be a separate PR.

adbar:
The files were what I meant, thanks!

PR #135, opened 2024-06-10
Feat/better dictionaries

juanjoDiaz:
Here is an alternative to #133 

It has several advantages:
* It encapsulates better the bytestring optimization while keeping the dictionary factory working on strings
* It does not require a external dependency to pick a temp folder for disk caching
* It does not require all the additional hashing logic. It simply caches data in a folder specific to simplema's version or wherever the user says
* It does not use internal functions of simplemma like `_load_dictionary_from_disk`

I still think that this would be better publish as a separate module and not as part of simplemma.


adbar:
@juanjoDiaz Now that #133 has been merged do you want to keep working on it or should we close this PR for now?

Issue #152, opened 2024-10-24
Remove `langdetect` submodule


adbar:
`langdetect` has been replaced by `language_detector`, the corresponding functions raise `ValueError`, the submodule can be removed in a future version.

PR #153, opened 2024-10-25
Refactor: Rename lang_detector() to langdetect()

shrijayan:
This PR renames the lang_detector() function to langdetect() for improved readability and alignment with standard naming conventions across the API.

Changes:
Renamed lang_detector() to langdetect().
Updated references in documentation and example code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,16 +197,16 @@ Language detection works by providing a text and tuple `lang` consisting
 of a series of languages of interest. Scores between 0 and 1 are
 returned.
 
-The `lang_detector()` function returns a list of language codes along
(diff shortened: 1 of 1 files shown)


adbar:
@shrijayan Thanks for the PR, the idea isn't wrong however right now API stability is the priority and I don't want to make a change just for the sake of it.

Please also note that tests would have to be updated and deprecation warnings set. The corresponding functions would also have to be removed at a later time in order to give users time to adapt, see issue #152. All in all it's not worth it.

So I'm refusing the PR, feel free to focus on the lemmatization algorithms or rules if you're interested.

PR #156, opened 2024-11-05
Fix data written by dictionary_pickler.py

Dunedan:
The changes to the return types of `DictionaryFactory.get_dictionary()` in 63933fc broke the generation of dictionaries using `training/dictionary_pickler.py`. This commit fixes that again, by undoing the changes to `training/dictionary_pickler.py` made by 63933fc.


adbar:
Neat, thanks!

PR #169, opened 2026-05-30
docs: review, prune and update

adbar:


docs/contributing.md
```diff
@@ -1,39 +1,40 @@
-## Contributions
+# Contributing
 
-### Issues, bugs & improvment ideas
+{%
+   include-markdown "../README.md"
+   start="<!-- include:contributions:start -->"
+   end="<!-- include:contributions:end -->"
+%}
```
juanjoDiaz, line 7: This block doesn't work in markdown.

simplemma/lemmatizer.py
```diff
@@ -3,7 +3,7 @@
 Provides classes for lemmatizing token and full texts.
 
 - [Lemmatizer][simplemma.lemmatizer.Lemmatizer]: Class for performing token and full text lemmatization.
-- [is_known()][simplemma.lemmatizer.is_known]: A legacy function that wraps the Lemmatizer's [is_known()][simplemma.lemmatizer.Lemmatizer.is_known] method.
+- [is_known()][simplemma.lemmatizer.is_known]: A legacy function that checks whether a token is present in the language data.
```
juanjoDiaz, line 6: I think that it was good to be clear on what it this legacy function wrapping so people know what they should be using instead.

juanjoDiaz:
@adbar not sure how you are viewing the docs but this PR break them in GitHub.

docs/contributing.md
adbar: It's for publication here: https://adbar.github.io/simplemma/contributing/

simplemma/lemmatizer.py
adbar: Sorry I missed that, we could open an issue or a PR

adbar:
Duplicate doc sections had diverged and mkdocs can work with the syntax. IMHO the main readme is fine and the deployed pages look good, if not please open an issue with the findings.

PR #171, opened 2026-06-11
fix: code hardening to prevent bugs

### LINKED ISSUE #55: Establish linting and quality tools
We should establish how we want to lint the code.
Currently this repo includes flake8, mypy, black and pytest.
You seem to also use pylint.

We should document the agreed approach and rules in the readme and also put them in the pipeline.

adbar:
- Correctness: NFC input normalization, prevent errors and add corresponding tests
- Security: add token-length cap
- Simplification: single-pass affix collapse, faster native .get() dictionary access, code simplification
- Maintenance: replace flake8 with ruff and adjust code

### DIFF
--- a/simplemma/strategies/greedy_dictionary_lookup.py
+++ b/simplemma/strategies/greedy_dictionary_lookup.py
@@ -10,6 +10,11 @@ from .lemmatization_strategy import LemmatizationStrategy
 SHORTER_GREEDY = {"bg", "et", "fi", "lv"}
 
 
+def greedy_min_length(lang: str) -> int:
+    """Shortest token worth decomposing; shorter ones are returned/skipped as-is."""
+    return 6 if lang in SHORTER_GREEDY else 8
+
+
 class GreedyDictionaryLookupStrategy(LemmatizationStrategy):
     """
     This class represents a lemmatization strategy that performs lemmatization using a greedy dictionary lookup strategy.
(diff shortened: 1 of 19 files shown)


simplemma/strategies/greedy_dictionary_lookup.py
```diff
@@ -10,6 +10,11 @@
 SHORTER_GREEDY = {"bg", "et", "fi", "lv"}
 
 
+def greedy_min_length(lang: str) -> int:
```
juanjoDiaz, line 13: should this be a static method of the `GreedyDictionaryLookupStrategy` to keep the class-based approach?

simplemma/strategies/hyphen_removal.py
```diff
@@ -53,8 +53,10 @@ def get_lemma(self, token: str, lang: str) -> str | None:
             str | None: The lemma for the token, or None if no lemma is found.
 
         """
+        if not any(hyphen in token for hyphen in HYPHENS):
```
juanjoDiaz, line 56: Is this faster than the regex?

juanjoDiaz (approved):
Lgtm!

Just couple of minor comments!

simplemma/strategies/greedy_dictionary_lookup.py
adbar: I think module-level placement is easier here.

simplemma/strategies/hyphen_removal.py
adbar: Yes, the most common case is that most tokens don't contain hyphens.

PR #172, opened 2026-06-11
maintenance: add langdetect safeguards

adbar:


simplemma/language_detector.py
```diff
@@ -22,7 +22,7 @@ def in_target_language(
     text: str,
     lang: str | tuple[str, ...],
     greedy: bool = False,
-    token_sampler: TokenSampler = MostCommonTokenSampler(),
+    token_sampler: TokenSampler | None = None,
```
juanjoDiaz: Why making `token_sampler` nullable, instead of just ensuring the right default here?

simplemma/language_detector.py
```diff
@@ -47,10 +49,7 @@ def langdetect(
     text: str,
     lang: str | tuple[str, ...],
     greedy: bool = False,
-    token_samplers: list[TokenSampler] = [
-        MostCommonTokenSampler(),
-        RelaxedMostCommonTokenSampler(),
-    ],
+    token_samplers: list[TokenSampler] | None = None,
```
juanjoDiaz, line 49: Same question as above

simplemma/language_detector.py
```diff
@@ -130,49 +130,39 @@ def __init__(
...
-        self._orig_token_sampler = token_sampler
         self._lemmatization_strategy = lemmatization_strategy
 
-    def _restore_token_sampler(self) -> None:
-        self._token_sampler = self._orig_token_sampler
-
     def proportion_in_each_language(
         self,
         text: str,
+        token_sampler: TokenSampler | None = None,
```
juanjoDiaz: Good catch about the concurrency issue!!

However, this change makes the API of the class `LanguageDetector` a bit weird.
The original idea is that you instantiate the class with a `token_sampler` and that's the token sampler that the instance will always use.

Maybe a cleaner solution could be to add `additional_token_samplers` to the class constructor or even make it possible for `token_sampler` to be a list, and use those in all the methods?

simplemma/language_detector.py
```diff
@@ -192,21 +182,39 @@ def proportion_in_target_languages(
...
-            for lang_code in self._lang:
-                candidate = self._lemmatization_strategy.get_lemma(token, lang_code)
-                if candidate is not None:
-                    in_target += 1
-                    break
-        return in_target / len(tokens)
+        _, found_any = self._count_known_tokens(tokens)
+        return found_any.count(True) / len(tokens)
+
+    def _count_known_tokens(
```
juanjoDiaz: I like the languagees-outer change for better caching.

I'm not sure if it's worth it to extract the code to a common method like this.
It cause a small performance and memory-consumption impact in both methods, especially in `proportion_in_target_languages` as the code does more things unnecessarily.
But I'm OK keeping it if you think that's better.

juanjoDiaz (commented):
A couple of comments from my side

simplemma/language_detector.py
adbar: For consistency reasons. A bit of an edge case but a mutable list was instantiated once at definition time. This new pattern is also used in in_target_language.

simplemma/language_detector.py
adbar: Parameter-passing is the only approach that is both stateless and concurrency-safe, it's slightly cleaner than the alternatives IMO.

simplemma/language_detector.py
adbar: Same issue as above, feel free to open an issue if you'd use a list for `token_sampler` in the future.

simplemma/language_detector.py
adbar: Yes, I'll see if I can find a simpler way of doing this

adbar:
@juanjoDiaz Thanks for the comments, they're now addressed.

PR #181, opened 2026-07-20
fix: add zero-dependency low-memory option

### LINKED ISSUE #72: Investigate other data structures to store language data

adbar:
Simplemma loads full dictionaries into RAM per language. This branch adds an opt-in low-memory path so users can trade RAM for speed:
- New `low_memory=True` flag on all the main functions (lemmatize, is_known, langdetect, etc.) plus DefaultStrategy. Picks the best available backend
-  New `StreamDictionaryFactory`: a zero-dependency backend reading dictionaries straight off disk instead of loading them into a dict (~50 MB vs ~175 MB RAM on German, at a speed cost)


Dunedan:
That's an interesting approach. I wonder how the three dictionaries factories compare with the new dictionary compression format in terms of loading time, memory usage and lookup performance. Maybe you can add some benchmarks later on. Might even be worth adding a table with different benchmarks to the README to make it easier for users to pick the most suitable dictionary factory for their use case.

